PR #5, opened 2018-08-16
Fix small typo

Novemser:
Replace duplicated "what what" with "on what".


quicquid:
I think one copy of "what" just needs to be deleted - i'm not sure about "on what".

lemmy:
Not a native speaker either but "on" sounds funny to me too.

Novemser:
Agree.

PR #14, opened 2020-03-26
adding a new toy specification of paxos

neoschizomer:
I'm writing a toy specification of paxos by using the inbox and outbox of each replica as its state. Actually I'm a new player of TLA+ and I'm not sure whether it is a good way.

Looking forward to your reply and thanks very much.


lemmy:
@neoschizomer Thanks for your effort.  Your spec has no comments, which makes it unsuited to be an example.  Could you add comments?

neoschizomer:
@lemmy thanks for your reply. I have added some comments into the specification.

neoschizomer:
Thanks a lot for your suggestions especially for the constant anti-patterns and I have updated the spec.

PR #17, opened 2020-08-11
Add specification for ViewStamped Replication

pkj415:


lemmy:
The PR doesn't update the top README.MD.

lemmy:
@muenchnerkindl Do you have time to review?

muenchnerkindl:
Sure, I'll review this.

muenchnerkindl:
Thanks for the specification: this is certainly an interesting algorithm for TLA+. I have a few suggestions for improving it (cf. comment lines "\* sm" in the attached TLA+ modules):

- update the top-level README.md to include this spec, as suggested by @lemmy,
- add a reference to the article that describes the algorithm in the header of the TLA+ module,
- the discussion of the fairness condition and the steps becoming disabled is a bit confused, and anyway there are no liveness properties stated in the modules,
- perhaps move the constraints on the state space size to model constraints rather than include them in the action definitions (that's what state constraints are for),
- there are two types of RECOVERYRESPONSE messages (those sent by the leader and by other nodes): I suggest to either disambiguate the message kind or to unify the fields these messages contain,
- the module contains a few TODOs, for which it is not always clear if they have been resolved or not,
- plus some suggestions here and there (and please check for typos).

Also, I was able to run TLC on a tiny model and verify the properties that appear in the module. It could however be useful to add a model to the repository that you consider of reasonable size.

None of these issues are mandatory but I'll close the PR for now. Please consider which suggestions you'd like to follow and open a new PR that I'll be happy to merge into the repository.


[LinearizableOrdering.txt](https://github.com/tlaplus/Examples/files/5125096/LinearizableOrdering.txt)
[ViewStampedReplication.txt](https://github.com/tlaplus/Examples/files/5125099/ViewStampedReplication.txt)

PR #18, opened 2020-09-22
Fixing None in Paxos

konnov:
I believe there is a typo in the line: `None == CHOOSE v : v \notin Ballot`. Theoretically, `CHOOSE v: v \notin Ballot` is allowed to return a value for `None` that belongs to `Values`, but is outside of `Ballot`. The value `None` is used together with values, not ballots. The undefined value for ballot is -1.


muenchnerkindl:
Hi Igor, you are right. And indeed, the version of Paxos that is distributed with TLAPS has the same definition that you propose. I'm merging your PR.

PR #20, opened 2020-10-15
Upload TLA+ specifications for ParallelRaft

HappyCS-Gu:
ParallelRaft is a variation of the Raft consensus protocol proposed in [_PolarFS:An ultra-low latency and failure resilient distributed file system for shared storage cloud database_.](https://www.vldb.org/pvldb/vol11/p1849-cao.pdf) We proposed a new consensus protocol ParallelRaft-CE based on ParallelRaft. We provided the formal specifications for ParallelRaft and  ParallelRaft-CE and  verified the refinement from ParallelRaft to Multi-Paxos and the correctness of ParallelRaft-CE.


lemmy:
Can add another PR to reference your example from https://github.com/tlaplus/Examples/blob/master/README.md ?

PR #29, opened 2021-03-05
Add readers-writers problem example

Isaac-DeFrain:


specifications/ReadersWriters/ReadersWriters.tla
```diff
@@ -0,0 +1,122 @@
...
+
+---------------------------------------------------------------------------
+
+(***********)
+(* Actions *)
+(***********)
+
+TryRead(actor) ==
+    /\ actor \notin WaitingToRead
+    /\ waiting' = Append(waiting, <<TRUE, actor>>)
```
muenchnerkindl, line 33: Since writing specifications is more about readability than (space) efficiency, I suggest adding tuples <<"read", actor>> and <<"write", actor>> and adapting the definitions of WaitingToRead and WaitingToWrite accordingly, i.e. something like

PendingReads == { req \in ToSet(waiting) : req[1] = "read" }
WaitingReaders == { req[2] : req \in PendingReads }

which I find easier to read (you may prefer to inline the definition of PendingReads but again, I find it more readable with two definitions). You'd also have to adapt the definition of ReadOrWrite.

specifications/ReadersWriters/ReadersWriters.tla
```diff
@@ -0,0 +1,122 @@
...
+    /\ UNCHANGED readers
+
+StopActivity(actor) ==
+    IF actor \in readers
+    THEN /\ readers' = readers \ {actor}
+         /\ UNCHANGED <<writers, waiting>>
+    ELSE /\ writers' = writers \ {actor}
+         /\ UNCHANGED <<readers, waiting>>
+
+ReadOrWrite ==
```
muenchnerkindl, line 59: This should be moved above the definition of StopActivity because it "belongs" to the definitions of Read and Write. Personally, I would make the definitions of the Read and Write actions self-contained, i.e. merge them with the relevant parts of ReadOrWrite. In particular, make all pre-conditions of the action explicit in the action: now, one has "writers = {}" in ReadOrWrite and then "readers = {}" in Write. But you may feel differently about this.

specifications/ReadersWriters/ReadersWriters.tla
```diff
@@ -0,0 +1,122 @@
...
+(**************)
+
+TypeOK ==
+    /\ readers \subseteq Actors
+    /\ writers \subseteq Actors
+    /\ waiting \in Seq(BOOLEAN \times Actors)
+
+Safety ==
+    /\ readers /= {} => writers = {}
+    /\ writers /= {} => readers = {}
```
muenchnerkindl, line 109: Note that the second conjunct is redundant with the first one since A => B is equivalent with ~B => ~A.
Perhaps writing "readers = {} \/ writers = {}" instead of these two conjuncts would be clearer.

specifications/ReadersWriters/ReadersWriters.tla
```diff
@@ -0,0 +1,122 @@
...
+CONSTANT NumActors
+
+VARIABLES
+    readers, \* set of processes currently reading
+    writers, \* set of processes currently writing
+    waiting  \* queue of processes waiting to access the resource
+
+vars == <<readers, writers, waiting>>
+
+Actors == 1..NumActors
```
muenchnerkindl, line 14: Why not make Actors a constant parameter of the module instead of requiring that set to be an interval of integers? Then you can make Actors a set of model values, say {alice, bob, charlie}, in the model.

muenchnerkindl (changes requested):
Please see the comments in the specification for suggestions on how to improve the spec. Also please add a README.md to the directory that explains what this spec does and what kinds of properties are verified.

Thanks,
Stephan

Isaac-DeFrain:
> Please see the comments in the specification for suggestions on how to improve the spec. Also please add a README.md to the directory that explains what this spec does and what kinds of properties are verified.
> 
> Thanks,
> Stephan

Thanks, I'll make some changes. There is a README.md, it's ignored by .gitignore for some reason.

specifications/ReadersWriters/ReadersWriters.tla
Isaac-DeFrain: I personally think it's better this way because you can easily change how many actors are in the model. I can make this change if you insist.

specifications/ReadersWriters/ReadersWriters.tla
Isaac-DeFrain: Point taken. Maybe it would be good to write it as: `~(readers /= {} /\ writers /= {})`

specifications/ReadersWriters/ReadersWriters.tla
```diff
@@ -0,0 +1,122 @@
...
+    /\ writers = {}
+    /\ waiting = <<>>
+
+Next ==
+    \/ \E actor \in Actors : TryRead(actor)
+    \/ \E actor \in Actors : TryWrite(actor)
+    \/ ReadOrWrite
+    \/ Stop
+
+Fairness ==
```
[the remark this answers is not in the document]
Isaac-DeFrain: Cool! I did not investigate that.

specifications/ReadersWriters/ReadersWriters.tla
Isaac-DeFrain: I doubt my "style" is idiomatic. I like this way (splitting up the action and enabling condition) better because, in my mind, then I can more easily think about these pieces separately.

I'll defer to your judgement here. It's not a big change...

specifications/ReadersWriters/ReadersWriters.tla
Isaac-DeFrain: This is valid. I'll change it.

specifications/ReadersWriters/ReadersWriters.tla
Isaac-DeFrain: After going to make this change, I didn't immediately see how to do it in a nice way. Since `waiting` is a queue and we need to inspect the head to know whether to trigger `Read` or `Write`, it's not clear how to make the actions self-contained without including an inspection of the head element of `waiting` in both. Maybe this is what you are suggesting, but I'd argue that that makes it harder to read. I'm curious what you would do in this case.

muenchnerkindl:
Thanks for your changes. In case it was not clear, my suggestions were only that, everybody has personal preferences.

Concerning fairness, why didn't you change the SF conditions for TryRead and TryWrite to WF? I just checked using TLC, and as expected, weak fairness is enough for ensuring the liveness property.

Regards,
Stephan

Isaac-DeFrain:
Oversight, my apologies.

muenchnerkindl:
Thank you for your contributions. I'll merge your PR into tlaplus/Examples and make the minor changes (in particular, weak instead of strong fairness) myself.

Regards,
Stephan

Isaac-DeFrain:
Thanks, @muenchnerkindl!

PR #35, opened 2021-05-16
[WIP] Read-Write quorum algorithm

sridhargopinath:
Hi,

I am new to TLA+ and am learning the PlusCal language. I found this repository very useful and wanted to contribute simple distributed algorithms to the repository that I am learning. I implemented the read-write quorum algorithm for operations on a shared variable. It is inspired by the below lecture.
https://www.youtube.com/watch?v=uNxl3BFcKSA

Please review and let me know of any changes I need to make.

Thanks!

### DIFF
--- a/specifications/read-write-quorum/Quorum.tla
+++ b/specifications/read-write-quorum/Quorum.tla
@@ -1,0 +1,274 @@
+------------------------------- MODULE Quorum -------------------------------
+
+EXTENDS Naturals, Sequences, FiniteSets, TLC
+
+CONSTANTS numClients, numReplicas, ReadQuorum, WriteQuorum, MaxClock
+
+\* TODO: Verify all uses of Pid
+Pid == 1 .. numClients+numReplicas
+ClientPids == 1 .. numClients
+ReplicaPids == numClients+1 .. numClients+numReplicas
+
+ReadQuorums == { x \in SUBSET ReplicaPids : Cardinality(x) = ReadQuorum }
+WriteQuorums == { x \in SUBSET ReplicaPids : Cardinality(x) = WriteQuorum }
+
+ClockVal == 0 .. MaxClock+1
+Message == [time: ClockVal, type: {"Read", "Write", "Ack"}, stamp: ClockVal, value: 1..20]
+
+(*--algorithm Quorum
+
+variables channel = [source \in Pid |-> [destination \in Pid |-> <<>>]]
+
+define
+  GetMsgSrcs(dst, type) ==
+    { src \in Pid: /\ Len(channel[src][dst]) > 0
+                   /\ Head(channel[src][dst]).type = type
+    }
+   
+  Max(a, b) == IF a <= b THEN b ELSE a
+end define
+
+macro Receive(type, clock, src, time, stamp, value) begin
+  with s \in GetMsgSrcs(self, type) do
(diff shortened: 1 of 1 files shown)


muenchnerkindl (changes requested):
Thanks you for your contribution! Follow a few suggestions for improvement:

1. Please add a comment explaining what the algorithm is supposed to achieve. It may also be worthwhile adding a few more correctness properties (invariants) to be verified. Finally, get read of the remaining "TODO" as this gives the impression of the module being unfinished.
2. Keep your specification as general as possible. In particular, the choice of 1..20 for values looks arbitrary. I suggest you introduce a parameter Value that can be instantiated for model checking. Similarly, I suggest you define ClockVal == Nat (since this is certainly what is intended in the algorithm) and override that definition only for model checking. (Actually, since it is not used anywhere, it should be enough to add a state constraint that ignores states for which the bound on the clock value is exceeded.)
3. The parameters dsts and clock of macro BroadcastTo are not used. The use of this macro suggests that the message is intended to be send only to (quorums of) replicas but the definition of macro sends the message to all other clients as well as to all replicas. Please make it clear what is intended. Similarly, the parameter clock is not used in macro SendTo.
4. Instead of using a single process type with different behavior for clients and replicas, define two process types (i.e., process Client \in ClientPids and process Replica \in ReplicaPids). This will result in a more readable algorithm.
5. Several variables (src, time, stamper, value) are used just to extract the components of a message in order to handle the message locally and are dead afterwards. These variables blow up the state space and may degrade the performance of TLC. I realize that the definition of View is intended to mitigate this problem. An IMHO cleaner way of achieving this is to write something like
  with src \in GetMsgSrc(...) do
     with msg = Head(channel[src][self]) do
       channel[src][self] := Tail(@);
       asks := Append(asks, [source |-> src, stamp |-> msg.stamper, ...])
     end with;
  end with
where src and msg are not state variables but just locally scoped identifiers. One drawback of this is that you have to inline the Receive macro but see below for an alternative.
6. The extra labels L2, L3, and L4 do not seem to be logically necessary but are there only because you cannot receive and send in the same atomic step as this would result in two assignments to the channel variable. You can get around this problem by modifying your channel operations so that they are "functional" rather than "imperative". More concretely, make them return the new channel value, e.g.
  define Send(ch, src, dst, msg) == [ch EXCEPT ![src][dst] = Append(@, msg)]
You can "chain" intermediate values of the channel through several network operations along the lines of
  with src \in GetMsgSrc(...) do
     with msg = Head(...), _chan = Tail(...) do
         clock := Max(clock, msg.time);
         chan := Send(_chan, self, src, [...])
     end with;
  end with
(Instead of inlining the Receive macro, you can define a similar Receive operator that returns both the new channel and the message that has been received. However, the guard ensuring that there must be a pending message cannot be put into a definition because operators must be total.)
Introducing coarser-grained atomic actions helps speed up model checking, and do not be afraid of letting TLC perform "expensive" local computations: these are anyway much cheaper than generating extra states.

I hope you find these suggestions helpful, and I'll be happy to accept your contribution if you address the above issues.

Stephan

sridhargopinath:
Thank you for the detailed list of changes. I will incorporate these changes and will update the PR.

lemmy:
@sridhargopinath Feel free to re-open.

PR #36, opened 2021-06-02
Refinement mapping for AsyncTerminationDetection => SyncTerminationDetection

lemmy:
@muenchnerkindl TLC "agrees" with this refinement mapping for `N = 3` and `N = 5`.  Do you too?  I stopped at the third level of a refinement proof.


muenchnerkindl:
@lemmy Yes, that's nice! I'll try to write a formal proof.

PR #38, opened 2021-08-11
Add CRDT-Bug and asyncio-lock examples

Alexander-N:
https://github.com/Alexander-N/tla-specs/issues/1


muenchnerkindl:
Thank you for your nice contributions! I am happy to merge your PR, but I'd suggest that you add a few comments to the specs illustrating the CRDT bug, for your own benefit and that of others.

Alexander-N:
Thanks for the feedback. Don't you feel that this would simply duplicate information from the readme? Since these are direct translations of the pseudocode all context is contained in the twitter thread. Maybe I can add some comments to make the mapping to the pseudocode more explicit?

muenchnerkindl:
Of course this was just a suggestion, and it is entirely up to you. I was mainly thinking of explaining the relationship between the pseudocode and the TLA spec, as well as highlighting the changes in the fixed version. In my experience, one entirely forgets about the details of the spec after a few months.

PR #40, opened 2021-08-27
Fix liveness checking in bcastByz

xosmig:
Similar to #39

There was a mistake in the fairness condition. It implicitly included the actions of Byzantine processes, which practically forces the Byzantine processes to cooperate with the correct processes.

To verify that there was a mistake, you can change the threshold in `UponAcceptNotSentBefore` and `UponAcceptSentBefore` from `N - T` to `N - 2*T`. This change breaks the protocol and the model checking should fail, but it will pass with the original specification due to a too strong fairness condition.


lemmy:
Is this ready to be merged?

banhday:
Thank for the fix! Since Receive is changed and two new formulas are introduced, TLAPS needs more additional user assistants. However, the main proof steps are unchanged since it is independent to the fairness condition. I'll fix it soon.

banhday:
I merged this change.

PR #44, opened 2021-10-31
Add Twirp Wire Protocol v7

mattmccormick:
This change adds a Twirp Wire Protocol (v7) TLA+ spec.  The protocol is defined at https://twitchtv.github.io/twirp/docs/spec_v7.html

I am new to TLA+ and we use Twirp at work so I thought this would be a good learning exercise for me.  I'm happy to receive any feedback on how to improve the spec.  If it's deemed too simple for inclusion in the Examples that's ok as well.


lemmy:
@mattmccormick Thanks for opening this pull request.  Given that the spec is small, readers will probably wonder what insights writing this spec provided.

specifications/Twirp/Twirp.tla
```diff
@@ -0,0 +1,72 @@
...
+Next == Success("") \/ Error
+
+Done == response.status # IncompleteStatus
+
+Spec == 
+    /\ Init 
+    /\ [][Next]_vars 
+    /\ WF_vars(Next) \* Use Weak Fairness to prevent stuttering since we can assume that some response will be returned even if the server crashes
+----
+Prop == <>[]Done
```
muenchnerkindl, line 71: This is rather trivially true, given that a single step achieves Done (and no other step is possible in that case).

Could you think of other invariants besides TypeOK?

muenchnerkindl (commented):
As @lemmy already pointed out, this is a really simple example. I add a few more detailed comments on TLA+ style.

mattmccormick:
> @mattmccormick Thanks for opening this pull request. Given that the spec is small, readers will probably wonder what insights writing this spec provided.

For me personally, as a newbie to TLA+, some of the things I got stuck with longer than I would have liked were:
- how to get the "values" from a function.  With `CodeToStatus` I wanted to be able to pull out the values like interacting with a `dict` in Python where I can do `d.values()`.  Finally, I was able to do that by using the DOMAIN which provides the `keys` and then getting the set of all values corresponding to all keys in the domain.
- in general, how to structure the spec and when to "pull out" actions.  I'm still not entirely clear on the best case practices as compared to normal programming where I have intuition based on experience.  In another spec I wrote, I seemed to struggle with pulling out actions that sometimes used prime variables and sometimes didn't.  I found it confusing to interleave actions that set the primed variables and ones that don't especially if the ones that set the primed variables do not set all the primed variables - so only a subset
- I have imported this module into another module and by doing so, I tried to make this more general.  That's why there are some formulas that aren't used here but are used in the other module.
- as mentioned in the comment, how much should actually be implemented in a TLA+ spec versus left out to avoid redundancy.  I decided to try to add everything I saw in the written specification but perhaps this was overboard and it's a good learning point to try to keep TLA+ specs simpler.

lemmy:
Quick question re 1) Were you aware of the [CommunityModules](https://github.com/tlaplus/CommunityModules) and the [`Range`](https://github.com/tlaplus/CommunityModules/blob/be80f411eb3b79972016fc73832f8f42ff1ac7de/modules/Functions.tla#L16-L21) operator specifically?  We've been considering graduating some operators/modules from the CommunityModules into TLC, and `Functions.tla` seems like the most useful one.

lemmy:
@mattmccormick Housekeeping, feel free to re-open if you still want this contributed.

Issue #46, opened 2022-04-28
Concurrent increments to a shared variable


Cjen1:
I've been using the example of [concurrently incrementing a shared variable](https://github.com/Cjen1/tla_increment) for demonstrating TLA+.

Since this can be split up into `t = i`, and `i = t + 1` actions, it demonstrates many of the interesting facets of TLA+ model checking while being a tiny example:
- Basic TLA+ syntax and TLC usage
- Non-determinism
- Counter-examples
- Liveness and fairness
- State space explosion and symmetry reduction

lemmy:
Thanks @Cjen1, can you open a PullRequest that adds your example to https://github.com/tlaplus/Examples/blob/master/README.md ?

PR #48, opened 2022-05-18
Add parallel increment example

Cjen1:
Adds the example described in issue #46.


muenchnerkindl:
Thank you for your contribution that I'm happy to merge into the list of examples. As a side remark, I do not really understand the motivation for the "fin" actions in the different specifications, given that TLA+ allows for stuttering steps in any case.

Cjen1:
@muenchnerkindl I think my intent was to model that each processes could continue taking steps, but I think that you are correct and an explicit action for that is unnecessary (I'll probably remove those steps at some point).

muenchnerkindl:
Thanks for your reply. As a general rule, we typically refrain from writing explicit stuttering actions. They are harmless for the next-state relation because they are subsumed by the possibility of stuttering that's baked into the TLA semantics. But they can create confusion when used in conjunction with fairness assumptions. For example, WF_vars(fin) expands to

```tla
(<>[] ENABLED <<fin>>_vars) => []<><<fin>>_vars
```

but this implication is vacuous since the action ```tla <<fin>>_vars``` is equivalent to FALSE and therefore never enabled.

PR #49, opened 2022-06-01
Rename read and write predicates in ReadersWriters specification

wrobell:
Use "IsRead/IsWrite" instead of "read/write" as names of the predicates
checking the type of processes.

Using "Is*" naming pattern for predicates in other specifications seems
to be the norm. It also makes ReadersWriters specification easier to
understand.


lemmy:
@Isaac-DeFrain Please comment given that you are the original author of the example.

lemmy:
Naming is hard.

PR #52, opened 2022-08-22
Fix an attribution and add Apalache column

nano-o:


muenchnerkindl (approved):
Thanks for making this more precise.

PR #54, opened 2022-09-28
adding PCR spec

egLanghaus:
this is mainly an excuse to use refinement mapping - the spec could be much better in terms of chemistry and biology

happy to make any needed changes!

### DIFF
--- a/specifications/README
+++ b/specifications/README
@@ -26,6 +26,9 @@ ewd840
    and van Gasteren for detecting distributed termination on a
    unidirectional ring, together with a safety proof.
 
+glowingRaccoon
+   A specification of PCR using refinement mapping in TLA+.
+
 lamport_mutex
    A TLA+ specification of the distributed mutual-exclusion
    algorithm that appeared as an example in Lamport's classic
--- a/specifications/glowingRaccoon/clean.tla
+++ b/specifications/glowingRaccoon/clean.tla
@@ -1,0 +1,73 @@
+----------------------------- MODULE clean -----------------------------
+(* what polymerase chain reaction is, and not how it works *)
+
+EXTENDS Naturals \* an import - copies module in there
+
+CONSTANTS DNA, PRIMER \* starting stock of key things
+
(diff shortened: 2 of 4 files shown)


lemmy:
I'm not familiar with PCR.  Would it make sense to document where chemistry and biology could be better?

muenchnerkindl:
... and in fact perhaps add a comment at the head of clean.tla for introducing the basic ideas to those of us who no nothing about the domain, plus similar comments at the head of the refining modules for describing the idea of those refinements?

But thanks for the contribution!

egLanghaus:
OK, I have made changes. For some reason, I was not able to add a README.md, so `clean.tla` has the necessary remarks.

muenchnerkindl:
Thank you very much for the explanations: I understand better what is going on.

A few more suggestions (none of this is mandatory, but you may want to consider them).
- Consider adding typing invariants to the specifications, for example (in clean.tla):

TypeOK ==
   /\ tee \in {"Warm", "Hot", "TooHot"}
   /\ primer \in Nat
   /\ dna \in Nat
   /\ template \in Nat
   /\ hybrid \in Nat

(This can actually replace the invariant `primerPositive'.)

- Consider adding more interesting invariants to your specifications. For example, in clean.tla you might add

preservation == template + primer + 2*(dna + hybrid) = PRIMER + 2 * DNA

which is an alternative (both more direct and stronger) way of expressing your action invariant `strandInvariant'.

- For both stages.tla and product.tla, you can replace strong fairness by weak fairness. In fact, no two actions are in conflict with another because the conditions on the unprimed variables are exclusive. It is usually considered good practice to have as weak specifications as possible.

egLanghaus:
Agreed! Changes pushed.

Good eye on the strong fairness - that was an artifact of an earlier version written before I hit upon "stage" as a way to drive the desired behavior.

muenchnerkindl:
Merging your contribution. Thank you again for adding it to the repository!

Issue #67, opened 2023-02-17
Specs failing proof validation


ahelwer:
I whipped up a quick script then downloaded & ran TLAPS 1.4.5 on macOS against all the modules containing proofs, and the following failed validation (no additional solvers installed, no additional command-line parameters given other than `-I` for include directory):
```
specifications/LoopInvariance/BinarySearch.tla
specifications/LoopInvariance/SumSequence.tla
specifications/Paxos/Consensus.tla
specifications/Paxos/Voting.tla
specifications/PaxosHowToWinATuringAward/Consensus.tla
specifications/PaxosHowToWinATuringAward/Voting.tla
specifications/ewd998/AsyncTerminationDetection_proof.tla
specifications/ewd998/EWD998_proof.tla
specifications/lamport_mutex/LamportMutex_proofs.tla
```
@muenchnerkindl I'd be interested in knowing how many of these are real proof failures and how many are due to wrong options or lack of installed solvers. My script was very simple, it just ran `tlapm` against each file and printed out the path if it returned a nonzero error code.

ahelwer:
I re-ran the script using the [202210041448](https://github.com/tlaplus/tlapm/releases/tag/202210041448) 1.5.0 pre-release and these are the list of failures instead:

Succeeded in 1.4.5 but fail in 1.5.0:
```
specifications/Bakery-Boulangerie/Bakery.tla
specifications/Bakery-Boulangerie/Boulanger.tla
```
Fail in both 1.4.5 and 1.5.0:
```
specifications/LoopInvariance/BinarySearch.tla
specifications/Paxos/Consensus.tla
specifications/Paxos/Voting.tla
specifications/PaxosHowToWinATuringAward/Consensus.tla
specifications/PaxosHowToWinATuringAward/Voting.tla
specifications/ewd998/EWD998_proof.tla
specifications/lamport_mutex/LamportMutex_proofs.tla
```
Upgrading to 1.5.0 fixed these proofs:
```
specifications/LoopInvariance/SumSequence.tla
specifications/ewd998/AsyncTerminationDetection_proof.tla
```

will62794:
[...]

PR #68, opened 2023-02-17
Added CI step to check proofs

ahelwer:
I think this is worth including so existing (and future) proofs will be checked. There are a number of proofs that are commented out in the script due to the following issues:
 - https://github.com/tlaplus/tlapm/issues/85
 - #67

Once these are fixed those comments can be removed. I added a step to cache the proof fingerprints which greatly speeds up checking for proofs that have already been checked.


ahelwer:
There is very high variance in how long it takes to check the proofs. Possible we could only check the ones that finish very quickly:
```
specifications/LoopInvariance/Quicksort.tla
Checked proofs in 58.1s
specifications/LoopInvariance/SumSequence.tla
Checked proofs in 308.6s
specifications/MisraReachability/ParReachProofs.tla
Checked proofs in 1.8s
specifications/MisraReachability/ReachabilityProofs.tla
Checked proofs in 11.9s
specifications/MisraReachability/ReachableProofs.tla
Checked proofs in 5.5s
specifications/TeachingConcurrency/Simple.tla
Checked proofs in 0.3s
specifications/TeachingConcurrency/SimpleRegular.tla
Checked proofs in 5.3s
specifications/TwoPhase/TLAPS.tla
Checked proofs in 0.3s
specifications/TwoPhase/TwoPhase.tla
Checked proofs in 0.3s
specifications/bcastByz/bcastByz.tla
Checked proofs in 266.6s
specifications/ewd840/EWD840_proof.tla
Checked proofs in 1.0s
specifications/ewd840/TLAPS.tla
Checked proofs in 0.3s
specifications/ewd998/AsyncTerminationDetection_proof.tla
Checked proofs in 1.3s
specifications/lamport_mutex/TLAPS.tla
Checked proofs in 0.3s
specifications/sums_even/TLAPS.tla
Checked proofs in 0.1s
specifications/sums_even/sums_even.tla
Checked proofs in 0.2s
```
I think they finish quickly when the fingerprints file is present, so another option is to cache the fingerprint files in between CI runs.

muenchnerkindl:
Not sure if it's a good idea to check proofs as part of a CI run. As you observed, proof checking can require considerable CPU time. We could indicate modules that have proofs that can be checked fast (similar to the "small models" that TLC is run on), but I'm not sure if it's worth the effort.

ahelwer:
@muenchnerkindl if I could cache the proof fingerprints would that be useful?

.github/workflows/CI.yml
```diff
@@ -110,6 +122,25 @@ jobs:
...
+      - name: Get proof fingerprint cache
+        if: matrix.os != 'windows-latest'
+        id: cache-proofs-restore
+        uses: actions/cache/restore@v3
+        with:
+          path: .tlacache
+          key: ${{ runner.os }}-proof-fingerprints
+      - name: Check proofs
+        if: matrix.os != 'windows-latest'
+        run: |
```
lemmy, line 134: Does the CI time out without old fingerprints? If not, the extra complexity might not be warranted.

lemmy (commented):
Let's keep the TLAPS.tla files to prevent parse errors for those users who do not care about proofs, unless the committed TLAPS.tla modules break the CI.

.github/workflows/CI.yml
ahelwer: Not currently, but once some of the currently-commented-out proofs are included the CI time will increase by quite a bit (probably 5-10 minutes).

lemmy: IMO it is better to catch regression at the expense of increased CI times, rather than to distributed broken examples.

PR #69, opened 2023-02-20
Fix module search paths for SANY and TLC

ahelwer:
Fixes current CI failures originally caused by adding dependency on TLAPS.tla to a model-checked module. TLC now finds TLAPS.tla. Switched from using `-DTLA-Library=$PATH` to `-cp tla2tools.jar:$PATH`.
See https://github.com/tlaplus/tlaplus/issues/795


ahelwer:
FYI @muenchnerkindl this fixes the CI. If you're making changes to this repo and are concerned they might cause CI failures (which will become more meaningful as these scripts mature and are battle-tested) you can open up a pull request like this, and the CI will be run against it so you can check it before pushing to the master branch.

PR #73, opened 2023-03-29
increase timeout for Z3 for one hard-to-prove obligation

damiendoligez:
Fixes one part of #67 : the failure on Bakery.tla.


muenchnerkindl:
Sure, go ahead. I find it just a little strange because on my machine this obligation is checked very quickly (using tlapm-1.5.0):

@!!BEGIN
@!!type:obligation
@!!id:2
@!!loc:398:5:398:7
@!!status:proved
@!!prover:z3
@!!meth:time-limit: 5; time-used: 0.5 (10%)
@!!already:false
@!!END

PR #74, opened 2023-04-04
Added three sets of specs

ahelwer:
- LeastCircularSubstring, a PlusCal spec of a complicated string algorithm
- LearnProofs, some very basic TLA⁺ proofs over PlusCal algorithms
- CheckpointCoordination, a spec I wrote while working inside Microsoft Azure Networking (now open-sourced with consent of Luis Irun-Briz)


specifications/CheckpointCoordination/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+This spec was written in June 2017 while the author worked in Microsoft Azure Networking, on the Azure DNS team.
+At the time the Azure DNS backend used an in-memory tree database called [RingMaster](https://github.com/Azure/RingMaster) built on top of a Paxos implementation called the [Replicated State Library (RSL)](https://github.com/Azure/RSL).
+Both of those systems were long ago released as open source, and now this spec has followed.
+This spec was written while designing a checkpoint/backup coordination feature described below.
+It is notable for two reasons:
+ 1. The TLC model checker found a 12-step error trace for a very subtle bug introduced by an attempted optimization
+ 2. The actual feature was implemented directly after the spec was written, providing an account of the experience of writing a program following a TLA⁺ spec
+
+Both of these points are covered in detail in [this](https://ahelwer.ca) blog post.
```
lemmy, line 11: There seems to be a missing link here.

ahelwer: Oh yeah I'm in the process of writing the post. When it's done I'll update the link the. mark this as ready to merge!

lemmy:
In case other reviewers also didn't know about draft pull requests: https://github.blog/2019-02-14-introducing-draft-pull-requests/

ahelwer:
@muenchnerkindl I'm interested in what you think of the proofs in AddTwo.tla and FindHighest.tla, especially the latter. I suspect I greatly complicated the process of going from inductive invariant to proving correctness.

muenchnerkindl:
I’m currently traveling without my laptop and will have a look when I’m back next week.

ahelwer:
@lemmy @muenchnerkindl well anyway these specs are all good to merge

specifications/LearnProofs/FindHighest.tla
```diff
@@ -0,0 +1,243 @@
...
+  <1>b. InductiveInvariant /\ UNCHANGED vars => InductiveInvariant'
+    BY DEFS InductiveInvariant, vars
+  \* Here we introduce TypeOK and TypeOK' as powerful assumptions, since we
+  \* have already proved that Spec => []TypeOK. Note that TypeOK is a
+  \* separate assumption from TypeOK' - including both can be the difference
+  \* between a proof being impossible or trivial!
+  <1>c. InductiveInvariant /\ TypeOK /\ TypeOK' /\ Next => InductiveInvariant'
+    <2>a. InductiveInvariant /\ Terminating => InductiveInvariant'
+      BY DEFS InductiveInvariant, Terminating, vars
+    <2>b. InductiveInvariant /\ TypeOK /\ TypeOK' /\ lb => InductiveInvariant'
```
muenchnerkindl, line 129: Somehow the inclusion of TypeOK' appears to confuse the provers. The following works:
     <2>b. InductiveInvariant /\ TypeOK /\ lb => InductiveInvariant'
       BY DEFS InductiveInvariant, TypeOK, lb, max

specifications/LearnProofs/FindHighest.tla
```diff
@@ -0,0 +1,243 @@
...
+THEOREM DoneIndexValueThm == Spec => []DoneIndexValue
+PROOF
+  <1>a. Init => DoneIndexValue
+    BY DEF Init, DoneIndexValue
+  <1>b. DoneIndexValue /\ UNCHANGED vars => DoneIndexValue'
+    BY DEFS DoneIndexValue, vars
+  <1>c. DoneIndexValue /\ TypeOK /\ Next => DoneIndexValue'
+    <2>a. DoneIndexValue /\ Terminating => DoneIndexValue'
+      BY DEFS DoneIndexValue, Terminating, vars
+    <2>b. DoneIndexValue /\ TypeOK /\ lb => DoneIndexValue'
```
muenchnerkindl, line 169: Again, the following proof works:
       BY DEFS DoneIndexValue, TypeOK, lb

specifications/LearnProofs/FindHighest.tla
```diff
@@ -0,0 +1,243 @@
...
+        /\ Next
+        => Correctness'
+    <2>a. Correctness /\ Terminating => Correctness'
+      BY DEF Correctness, Terminating, vars
+    <2>b.
+        /\ Correctness
+        /\ InductiveInvariant'
+        /\ DoneIndexValue'
+        /\ lb
+        => Correctness'
```
muenchnerkindl, line 212: Here as well, the following simple proof works:
       BY DEF Correctness, InductiveInvariant, DoneIndexValue, lb

muenchnerkindl (commented):
@ahelwer: Apologies for letting your message slip in my email stack. I finally had a look at your proofs: they are fine, but they can also be simplified quite a bit. I don't really know if you wanted to explore the technique of hierarchical subproofs, but they are hardly necessary for these specs. Note that it regularly happens when you are starting to prove some assertion that you don't really know what facts to use and which definitions to expand. So you'll start and write deep proofs only to see that in the end you could have done your proof in a single line. I usually find shorter proofs preferable because they are easier to digest. But it's your call which version you want to show.

specifications/LearnProofs/AddTwo.tla
```diff
@@ -0,0 +1,85 @@
...
+\* END TRANSLATION 
+
+TypeOK == x \in Nat
+
+THEOREM TypeInvariant == Spec => []TypeOK
+  <1>a. Init => TypeOK
+    BY DEF Init, TypeOK
+  <1>b. TypeOK /\ UNCHANGED vars => TypeOK'
+    BY DEF TypeOK, vars
+  <1>c. TypeOK /\ Next => TypeOK'
```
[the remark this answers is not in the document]
ahelwer: Nope, thought it was necessary but I suppose not. Thanks!

specifications/LearnProofs/AddTwo.tla
```diff
@@ -0,0 +1,85 @@
...
+a|b == \E c \in Nat : a*c = b
+
+AlwaysEven == 2|x
+
+THEOREM Spec => []AlwaysEven
+  <1>a. Init => AlwaysEven
+    BY DEF Init, AlwaysEven, |
+  <1>b. AlwaysEven /\ UNCHANGED vars => AlwaysEven'
+    BY DEF AlwaysEven, vars
+  <1>c. AlwaysEven /\ Next => AlwaysEven'
```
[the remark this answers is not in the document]
ahelwer: Wow, very impressive! How'd you come up with that second faster way of proving it?

ahelwer:
Thanks @muenchnerkindl! Wow I have a lot to learn about the proof system, I don't know how I managed to miss those quick proofs and go down the arithmetic proof rabbit holds.

specifications/LearnProofs/AddTwo.tla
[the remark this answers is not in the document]
muenchnerkindl: Well, it's actually not so hard if you have a basic understanding of how automatic theorem provers work. The issue here is that we have a combination of arithmetic reasoning and quantifier instantiation. From the assumption and the definition of the action we obtain something like x' = 2*c + 2 (for some c \in Nat), but we need to prove that x' = 2*d, for some d \in Nat. For humans, this is obvious, but automatic theorem provers use heuristics to find witnesses for existential quantifiers, and these heuristics will usually only try terms that appear in the context. In this case, there is no term 2*d available to the prover that fits the bill. The additional fact serves to introduce the required term 2*(c+1), and an SMT solver can easily prove that 2*c + 2 = 2*(c+1) and that c+1 \in Nat if c \in Nat. For some reason, Z3 would still not use that term to instantiate the quantifier (perhaps CVC4 or CVC5 would have done so), but Zenon is much more reliable when it comes to quantifier reasoning, so the combination of the two solves the goal. Not pretty, but doing quantifier instantiation modulo arithmetic equality reasoning is actually really hard.

muenchnerkindl (approved):

PR #76, opened 2023-04-12
Add Simplified Fast Paxos

TypeDefinition:


lemmy:
Thanks for the contribution.  In broad strokes, how is this version of FastPaxos different compared to the existing one?

TypeDefinition:
This example is greatly simplified as compared to Leslie Lamport's original paper by adding a few constraints.

We assume one unique coordinator in the system, rather than electing a new coordinator for each round. Thus there is no need for P1a and P1b messages and they can be omitted.

We also assume that every node in the system can communicate with one another, so there is no need to model _goodSets_.

Resolving value collisions has also been simplified by using the value selection rules as mentioned in Zhao Wenbing's Fast _Paxos Made Easy_ paper.

1. If no acceptor has casted any vote, then the coordinator is free to select any value for phase 2;
2. If at most a single value is present in the votes, then the coordinator must select that value;
3. If the votes contain different values, a value must be selected if the majority of acceptors in the quorum have casted a vote for that value. Otherwise, the coordinator is free to select any value.

lemmy:
It sounds like your spec should *refine* the original spec, which, if true, would be nice to formally establish.

TypeDefinition:
May I know how I can formally establish this? Apologies, I'm not too familiar with the process. Thank you!

lemmy:
See section 10.8 of Specifying Systems: https://lamport.azurewebsites.net/tla/book-02-08-08.pdf

TypeDefinition:
Thank you very much for the feedback. I'll close the PR for now until I am able to refine it.

lemmy:
Please note that my comment wasn't a rejection.

TypeDefinition:
Sorry, I misunderstood.

TypeDefinition:
May I clarify that it seems like in order to refine the original specification, I would need to have the original specification implemented, and reimplement mine on top of that? Whereas what I have currently is an implementation which is not an instance of original specification.

muenchnerkindl:
I don't really understand what you mean when you say that you "would need to have the original specification implemented" (as an executable program?). In TLA+, refinement means that every behavior allowed by your specification also satisfies the original FastPaxos spec. However, I can see that this is probably not the case given that you remove certain types of messages (whose presence is the pre-condition for actions of the original spec to happen). One could still hope to prove refinement under a refinement mapping where you show how to compute the state (and in particular the set of messages) of the original FastPaxos spec from the state of your spec. Just as @lemmy, I would not consider this as "nice to have" but not a prerequisite for adding your spec to the repository.

Technically, note that we require a DCO (developer certificate of origin) to be included in commits to this repo, see for example https://github.com/apps/dco, otherwise the CI will complain.

TypeDefinition:
Thanks for the clarification. In that case, I would like to merge this commit as-is if possible. I have also added the DCO. May I ask if the Authors section of the README.md table refers to the author of the research paper (Leslie Lamport), or the one who wrote the TLA+ code that is being submitted?

muenchnerkindl:
It's the author of the specification that should be given in the README.md.

TypeDefinition:
> It's the author of the specification that should be given in the README.md.

Thanks for the clarification.

TypeDefinition:
Hi, may I know if this can be merged? Thank you.

lemmy:
LGTM

specifications/SimplifiedFastPaxos/FastPaxos.tla
```diff
@@ -0,0 +1,170 @@
...
+    Same as in Paxos.
+*)
+ClassicAccepted ==
+    /\ UNCHANGED<<cValue>>
+    /\ PaxosAccepted
+
+(*
+    Same as in Paxos.
+*)
+ClassicDecide ==
```
muenchnerkindl, line 134: It would have been a little nicer if this definition had been written as
/\ UNCHANGED cValue
/\ PaxosDecide
which, I believe, is equivalent, but makes the correspondence more explicit.

specifications/SimplifiedFastPaxos/FastPaxos.cfg
```diff
@@ -0,0 +1,21 @@
...
+    v3 = v3
+    Values = {v1, v2, v3}
+    Ballots = {0, 1, 2, 3, 4}
+    FastBallots = {1, 3}
+    Quorums = {{r1, r2, r3}, {r1, r2, r4}, {r1, r3, r4}, {r2, r3, r4}}
+    FastQuorums = {{r1, r2, r3}, {r1, r2, r4}, {r1, r3, r4}, {r2, r3, r4}}
+
+SPECIFICATION FastSpec
+INVARIANTS FastTypeOK
+PROPERTY FastSafetyProperty
```
muenchnerkindl, line 20: Can you also check PaxosSpec, as announced in a theorem in FastPaxos.tla?

specifications/SimplifiedFastPaxos/Paxos.tla
```diff
@@ -0,0 +1,208 @@
...
+    /\ \/ decision = none
+       \/ \E m \in p2aMessages : m.value = decision
+    /\ \A m \in p1bMessages : /\ m.maxValue \in Values \/ 0 = m.maxVBallot
+                              /\ m.maxValue = none \/ 0 < m.maxVBallot
+
+\* At most 1 value can be learnt.
+PaxosConsistency == decision = none \/ decision = decision'
+
+PaxosSafetyProperty == /\ [][PaxosNontriviality]_<<messages, decision, maxBallot, maxVBallot, maxValue>>
+                       /\ [][PaxosConsistency]_<<messages, decision, maxBallot, maxVBallot, maxValue>>
```
muenchnerkindl, line 204: Since PaxosNontriviality is a state predicate, it can be checked as an invariant rather than as a temporal property.
Also, the second conjunct can be written a little more concisely as [][decision = none]_<<decision>>, which asserts that a decision can be taken only if its current value is none.

specifications/SimplifiedFastPaxos/Paxos.tla
```diff
@@ -0,0 +1,208 @@
...
+
```
muenchnerkindl, line 205: Given that PaxosSpec includes a fairness property, it would be nice to also check a liveness property if possible.

muenchnerkindl (commented):
Apologies for the delay. From my point of view, this example can be merged into the repository. I added a few comments that you may want to consider and that should not take you long to decide if you want to adapt your modules or leave them as they are.

specifications/SimplifiedFastPaxos/Paxos.tla
muenchnerkindl: Ah, Github pretty-printing ... That formula was
```tla
[][decision = none]_<< decision >>
```

specifications/SimplifiedFastPaxos/FastPaxos.cfg
TypeDefinition: May I know how I can do this? Is there a way to add 2 specifications to 1 .cfg file?

specifications/SimplifiedFastPaxos/Paxos.tla
TypeDefinition: I've added a reason for the lack of liveness property.

specifications/SimplifiedFastPaxos/FastPaxos.cfg
TypeDefinition: Thank you. I have removed the theorem.

PR #78, opened 2023-06-03
Parameterize parse/check python scripts for ease of use from other repos

### LINKED ISSUE #77: Parameterize version of TLA+ tools used during CI run
@lemmy in an earlier issue you mentioned it would be useful to parameterize the version of the TLA+ tools used during the CI run, so you could for example trigger a build of the tlaplus/tlaplus repo and have the resulting binary be used during a CI run for the tlaplus/examples repo. Are you still interested in this feature? Will look into implementing it if so.

ahelwer:
No longer do these scripts rely on various things being present at undocumented paths; all dependencies (tla2tools.jar path, manifest.json path, etc.) are now exposed using command line parameters. The scripts can be run from any working directory. These scripts also don't have any python package dependencies (tree-sitter etc.) so will be easy to run from the tlaplus/tlaplus repo CI instead of run.sh.

Also added a trigger to run the CI on a repository dispatch event, for the nightly tlaplus build to trigger.

Small amount of work necessary on tlaplus/tlaplus side but this is sufficient to close #77

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1335,8 +1335,8 @@
       "description": "Spec of simplified Fast Paxos from Lamport's 2006 paper Fast Paxos",
       "source": "",
       "authors": [
-        "Lim Ngian Xin Terry",
-        "Gaurav Gandhi"
+        "Gaurav Gandhi",
+        "Lim Ngian Xin Terry"
       ],
       "tags": [],
       "modules": [
--- a/.github/scripts/check_small_models.py
+++ b/.github/scripts/check_small_models.py
@@ -4,7 +4,9 @@ models should finish executing in less than ten seconds on the GitHub CI
 machines.
 """
 
+from argparse import ArgumentParser
 import logging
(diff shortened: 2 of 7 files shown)


manifest.json
```diff
@@ -1335,8 +1335,8 @@
       "description": "Spec of simplified Fast Paxos from Lamport's 2006 paper Fast Paxos",
       "source": "",
       "authors": [
-        "Lim Ngian Xin Terry",
-        "Gaurav Gandhi"
+        "Gaurav Gandhi",
+        "Lim Ngian Xin Terry"
```
lemmy, line 1339: Is this reordering intentional?

lemmy (approved):

manifest.json
ahelwer: I just re-ran the manifest generation script which orders authors alphabetically.

lemmy: Alphabetical order might not always be desirable.

Issue #80, opened 2023-06-28
Question about the specification of reliable broadcast algorithm by Bracha & Toueg (1985)


lduranovic99:
I'm confused about some of the choices made when the specification for this algorithm was encoded in TLA+. If I'm understanding things correctly, `sendEcho()` method for example seems to be prioritizing hosts that "start" in the `v1` location, as if `1` is always the value proposed by the transmitter. In the original paper, before the algorithm even starts, the transmitter (which is effectively the leader) is supposed to send out *initial* messages. In your TLA+ spec, these messages don't even exist. Why is this the case, and how do I convince myself that this simplification, or rewrite, is still safe and equivalent to the original algorithm?

My other concern is about the Byzantine processes. My intuition about this bit of your specification is that their behavior is "abstracted away", and that all the progress that the non-faulty hosts in the protocol are making is due to correct messages being sent and received. Although this makes sense to me on a high level, I'm still not seeing why the TLA+ spec is not encoding in some way the ability of faulty processes to send, for example, *echo* messages with differing values.

lemmy:
Tagging the spec authors: @konnov @banhday @josef-widder

lduranovic99:
@lemmy Thanks!

josef-widder:
Hi @lduranovic99 , could you please clarify which spec you are referring to. I guess there might be several and the encoding of faults may differ, so to answer your second question, I want to be sure we are on the same page ;-) 

Regarding the first concern, this encoding originated from our previous work on parameterized model checking threshold automata, e.g. [here](https://lmcs.episciences.org/10824). This method only works on symmetric systems, so we couldn't capture a special process (the broadcaster) in an efficient manner. At the same time, we wanted to verify the threshold functionalities (t+1, n-t under resilience conditions like n>3t). Thus, we have chosen to capture the effect of the broadcaster in the initial state:
- all processes have 0: e.g., the broadcaster is correct and didn't choose to send any message. 
- all processes have 1: e.g, the broadcaster is correct and thus sends to everyone
- mixed inputs: the broadcaster is faulty and sense to some (those with 1)

(Of course the "all 0s" and "all 1s" also capture the scenario where the broadcaster is "faulty", but sends or does not send consistently to everyone.)

From our initial states, a process sends echo if and only if it sends echo in executions the corresponding execution with a broadcaster. To formalize the argument, I guess I would define projections of traces that abstract away the broadcaster, and show that there is trace equivalence to the traces of the spec.

[...]

PR #81, opened 2023-07-03
Bug fix in aba_asyn_byz.tla

josef-widder:
Following the discussion in https://github.com/tlaplus/Examples/issues/80 I discovered a bug in the encoding w.r.t handling of messages by faulty processes. If a process was to become faulty after having sent messages, its messages might be counted twice, while the original algorithm counts messages by distinct senders, that is, it counts senders rather than counting messages.

The proposed change is the minimal I could figure out to fix that. Processes can turn Byzantine only in an initial process state. @konnov, can you also please have a look?

### DIFF
--- a/specifications/aba-asyn-byz/aba_asyn_byz.tla
+++ b/specifications/aba-asyn-byz/aba_asyn_byz.tla
@@ -59,9 +59,13 @@ Init1 ==
   /\ nByz = 0
   /\ pc \in [ i \in Proc |-> "V1" ]  
   
-(* If there are less than F Byzantine processes, process i becomes faulty.*)
+(* If there are less than F Byzantine processes, process i becomes faulty. *)
+(* We requite i to be in an initial state (V0 or V1) to not break the      *)
+(* message counting abstraction.                                           *)
 BecomeByzantine(i) ==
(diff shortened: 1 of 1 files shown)


konnov:
I agree that this behavior of `BecomeByzantine` does not look right. However, I have run TLC against `Unforg_Ltl`, `Corr_Ltl`, `Agreement_Ltl` for `N=4`, `T=1`, `F=1`. TLC reported no error. It would be great to understand why this specification actually works, though we both believe it should not.

konnov:
As we discussed with @josef-widder, the current specification seems to be an abstraction of the original algorithm. It does satisfy all three properties, but it has a larger set of behaviours than the original algorithm. The proposed fix brings the specification closer to the original algorithm.

konnov (approved):
Thank you for fixing this specification.

konnov:
Hi @lemmy! Could you merge this one?

muenchnerkindl:
Hi, following the requirements of the Linux Foundation, you need to include a DCO in your commit, e.g. https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md.

Thanks!
Stephan

PR #83, opened 2023-07-27
Fix tar CI failure on windows

ahelwer:
I think the lesson here is don't try to use tar on windows. Should probably do a squash-merge because a lot of these commits were small changes trying to get the CI to pass.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -16,16 +16,16 @@ jobs:
     strategy:
       matrix:
         include:
-        - { os: windows-latest, shell: bash }
-        - { os: ubuntu-latest,  shell: bash  }
-        - { os: macos-latest,   shell: bash  }
+        - { os: windows-latest }
+        - { os: ubuntu-latest }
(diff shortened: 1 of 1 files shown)


ahelwer:
@lemmy what do you make of this failure in the windows CI?
```
Error: TLC threw an unexpected exception.
This was probably caused by an error in the spec or model.
See the User Output or TLC Console for clues to what happened.
The exception was a java.lang.UnsupportedClassVersionError
: org/jungrapht/visualization/layout/algorithms/LayoutAlgorithm has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

lemmy:
TLC should be running with at least Java 11.

lemmy:
Can this PR be merged or is the class version mismatch an issue?

ahelwer:
I think this can be merged. It's a bit worrying to have a transient java version issue but I added a line to print out the java version to the CI so at least if it happens again we'll have a lead on why.

PR #84, opened 2023-08-03
Added Pluscal version of the key-value store with snapshot isolation,…

muratdem:
… instantiated clientcentric checking to model check for snapshot isolation properly


specifications/KeyValueStore/Util.tla
```diff
@@ -0,0 +1,49 @@
+-------------------------------- MODULE Util --------------------------------
+EXTENDS Sequences, FiniteSets, Naturals, TLC
+\* Simple utility functions
+intersects(a, b) == (Cardinality(a \cap b) > 0)
+max(s) == CHOOSE i \in s : (~\E j \in s : j > i)
+min(s) == CHOOSE i \in s : (~\E j \in s : j < i)
+
+\* Utilies from Practical TLA+
```
lemmy, line 8: The source of Practical TLA+ doesn't come with a [permissive license](https://github.com/Apress/practical-tla-plus/blob/master/LICENSE.txt).  The same or similar operators should exists in the CommunityModules.

muratdem: Util.tla came from/with ClientCentric.tla (which has a [permissive license](https://github.com/cwi-swat/tla-ci/blob/master/LICENSE))

It seems like only lines from 8-18 is implied to come from Practical TLA+. 
Would you be able to substitute with lines from community modules?

lemmy: https://github.com/tlaplus/CommunityModules/blob/master/modules/FiniteSetsExt.tla and https://github.com/tlaplus/CommunityModules/blob/master/modules/SequencesExt.tla appear to have similar operator definitions.

You should add a BSD license file in this folder since the examples repo is MIT licensed.

muratdem: Tried to fix things as much as I could, but I run into problems. So I am giving up.

muratdem: Ok, rewrote the three operators using examples from Community Modules, and added the license for the ClientCentric.tla

Should be ready to merge.

muratdem: Of course had to mention that Util now depends on Functions from Community modules. Arggh..

specifications/KeyValueStore/ClientCentric.tla
```diff
@@ -0,0 +1,185 @@
...
+\* Note: This execution does therefor not contain the "final state" of the execution, since it is not a parent state of a transaction.
+Execution == Seq(ExecutionElem)
+
+\* Seq
+executionStates(execution) == [ i \in 1..Len(execution) |-> execution[i].parentState ]
+\* Set
+executionTransactions(execution) == { ep.transaction : ep \in SeqToSet(execution) }
+
+\* "The parent state is the last state in the execution
+\* Definition 1: s -T-> s' ≡ [(k,v) ∈ s ∧ (k,v)􏰍 \notin s'] => k ∈ W_T /\ w(k,v) ∈ Σ_T => (k,v) ∈ s.
```
lemmy, line 43: I can't tell if the box `[]` after `(k,v)` is a unicode rendering issue or part of the notation.

muratdem: Looks like it is a typo and unicode rendering issue.
https://github.com/cwi-swat/tla-ci/blob/53390b453ebbe6ed79a3f751863f5a4cdf32c5a0/ClientCentric.tla#L43C48-L43C50

It is in a comment, so it shouldn't matter. Feel free to remove it before merging.

lemmy:
LGTM, @muenchnerkindl are you also going to review?

specifications/KeyValueStore/ClientCentric.tla
```diff
@@ -0,0 +1,185 @@
...
+
+PrereadAll(execution, transactions) == 
+  \A transaction \in transactions: Preread(execution, transaction)
+
+\* A state `s` is complete for `T` in `e` if every operation in `T` can read from `s`
+Complete(execution, transaction, state) == 
+  LET setOfAllReadStatesOfOperation(transactionOperations) ==
+        { ReadStates(execution, operation, transaction) : operation \in SeqToSet(transactionOperations) }
+     \* also include all states for when the transaction contains no operations
+      readStatesForEmptyTransaction == { s \in SeqToSet(executionStates(execution)) : beforeOrEqualInExecution(execution, s, parentState(execution, transaction)) }
```
muenchnerkindl, line 85: I don't see how this definition (or its use) is restricted to the case "when the transaction contains no operation"? Perhaps rephrase the comment?

specifications/KeyValueStore/Util.tla
```diff
@@ -0,0 +1,45 @@
+-------------------------------- MODULE Util --------------------------------
+EXTENDS Sequences, FiniteSets, Functions, Naturals, TLC
+\* Simple utility functions
+intersects(a, b) == (Cardinality(a \cap b) > 0)
```
muenchnerkindl, line 4: Defining this as "a \cap b # {}" would remove the dependence on FiniteSets.

muenchnerkindl (commented):
Looks interesting! Added a few suggestions, but none of them are mandatory. I let you review them before merging.

specifications/KeyValueStore/Util.tla
muratdem: OK, I will try this.

specifications/KeyValueStore/KVsnap.tla
```diff
@@ -0,0 +1,203 @@
...
+
+variables 
+    \* A data store mapping keys to values
+    store = [k \in Key |-> NoVal],
+
+    \* The set of open snapshot transactions
+    tx = {},
+
+    \* Snapshots of the store for each transaction
+    snapshotStore = [t \in TxId |-> [k \in Key |-> NoVal]],
```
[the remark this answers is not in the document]
muratdem: Let me try this.

specifications/KeyValueStore/ClientCentric.tla
```diff
@@ -0,0 +1,185 @@
+--------------------------- MODULE ClientCentric ---------------------------
+EXTENDS Naturals, TLC, Sequences, FiniteSets, Util
+VARIABLES Keys, Values
```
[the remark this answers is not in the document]
muratdem: ClientCentric is not my module, written by Tim, so I am reluctant to make any change there.

specifications/KeyValueStore/ClientCentric.tla
muratdem: ClientCentric is not my module, written by Tim, so I am reluctant to make any change there.

specifications/KeyValueStore/KVsnap.tla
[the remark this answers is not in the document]
muratdem: Ah I see. I prefer to include the invariants in the define block. Newcomers to TLA+/Pluscal, get confused about why the invariant conditions are missing, if they are listed after the generated TLA+, at the end of the file. So I like to list them more prevalently in the define block early on.

muratdem:
I don't understand why the download TLA+ dependencies step in CI is failing.
It is not due to the two line change I did in Util.tla, no?
>intersects(a, b) == a \cap b # {}
> take off FiniteSets from extends

muenchnerkindl:
The failures do not appear to be related to your spec: the Linux issue is a failure to build TLAPS (an issue that pops up occasionally, we don't really understand why), the MacOS build fails to what appears to be a version issue with the tree sitter grammar, which I have never seen before. Hopefully @lemmy and/or @ahelwer will be able to comment.

ahelwer:
Yeah this is a pretty weird error. Looking into it now.

specifications/KeyValueStore/ClientCentric.tla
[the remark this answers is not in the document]
lemmy: /cc @TimSoethout

specifications/KeyValueStore/ClientCentric.tla
lemmy: /cc @TimSoethout

TimSoethout:
Really cool to see that you are adding examples based on my earlier work. :)
Regarding the /cc mentions: I will have a look after my holidays, hopefully somewhere next week.

specifications/KeyValueStore/ClientCentric.tla
TimSoethout: `􏰍` seems to be a copy/paste artifact from the original CI papers PDF. It can be removed.

specifications/KeyValueStore/ClientCentric.tla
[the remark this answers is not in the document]
TimSoethout: The module is instantiated as seen in: https://github.com/tlaplus/Examples/pull/84/files/adb3d2aa3f62812a2becd5c2f6700d69eb1c5134#diff-ccf178cee4a4265b5f8405a78d596e79973a841e20119a281c19775a1ecae139R14
`CC == INSTANCE ClientCentric WITH Keys <- Key, Values <- TxId \union {NoVal} `

If something similar is possible with constant parameters, and it is more idiomatic TLA, you should definitely go for parameters.

specifications/KeyValueStore/ClientCentric.tla
TimSoethout: If I remember correctly, this is where I deviated from the original definitions in the paper by Crooks (https://www.cs.cornell.edu/lorenzo/papers/Crooks17Seeing.pdf). The comment is mainly there to upgrade the definition to also allow transactions with an empty operation set, while still satisfying the isolation level. 
I personally used when incrementally model checking isolation levels of the Two-Phase Locking and Local-Coordination Avoidance protocols, which contained empty transactions when initializing.
The comment could for example be rephrased to: 
"`readStatesForEmptyTransaction` contains all previous states, in order to make sure that empty transactions do not incorrectly invalidate the checked isolation level."

specifications/KeyValueStore/ClientCentric.tla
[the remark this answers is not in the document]
muenchnerkindl: Instantiation is certainly possible for constant as well as variable parameters. Declaring Keys and Values to be variables allows these sets to change during an execution, which might be interesting for some applications, but I don't know what the original paper had in mind. If they are constants, the module is a constant module (since all operator definitions are of constant level as well), which allows operators to be instantiated at constant, variable or action level. Potential complications with non-constant modules are described in section 17.8 of Specifying Systems. However, my comment was just a caveat for later, there is nothing wrong with the modules as they are written, and they are accepted by TLC.

specifications/KeyValueStore/ClientCentric.tla
lemmy: Has this conversation been resolved?  If yes, it looks like this PR is ready to be merged.

muenchnerkindl: I certainly have no objection to merging the PR, although I don't think that the wording of the comment has been changed (Tim's suggestion clarifies the intent of the definition IMHO).

lemmy: @muratdem can you do the finishing touches (wording of this comment + removal `[]` copy/paste artifact), and the we merge it?

muratdem: yes, I will also pull some global variables to be process local.

specifications/KeyValueStore/KVsnap.tla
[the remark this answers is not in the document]
muratdem: ok, pulled snapshotStore to be local, as it makes the model look easier to understand. The right thing to do. 
I got rid of the "written" as global variable: the in-process "write_keys" variable does the same work, so there is no need. This made the model checking even faster, cut half the time.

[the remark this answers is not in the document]
muratdem: I did prune the TypeOK to not mention SnapshotStore. This is a reasonable tradeoff, that is assigned to `store` and we are typechecking `store` anyways.

PR #89, opened 2023-08-23
Retry tlapm install on failure

ahelwer:
Workaround to reduce transient CI failures until this issue is fixed: https://github.com/tlaplus/tlapm/issues/88


lemmy:
The `run.sh` script might not be used by the CI, but there's no harm in keeping it. I occasionally use it locally.

PR #90, opened 2023-08-25
Configure TLAPM install to proceed if exiting with nonzero error code

ahelwer:
Github actions appears to run `set -e` by default, so the tlapm install failure will abort the script and ignore the retry logic around it. We counteract this by running `set +e` right before the tlapm install retry loop.


ahelwer:
@lemmy minor fix

PR #93, opened 2023-09-04
Majority

muenchnerkindl:
Added a new beginner-friendly TLA+ spec for an efficient majority vote algorithm due to Boyer and Moore. Comes with a model suitable for TLC and a proof that can be checked using TLAPS.


lemmy:
LGTM, but do you want the TLAPS.tla module to be checked in as well?

muenchnerkindl:
If it can be avoided, I’d rather leave it out but looking at other entries I got the impression that it needed to be present in order to avoid CI error messages?On 4 Sep 2023, at 20:37, Markus Alexander Kuppe ***@***.***> wrote:﻿
LGTM, but do you want the TLAPS.tla module to be checked in as well?

lemmy:
> If it can be avoided, I’d rather leave it out but looking at other entries I got the impression that it needed to be present in order to avoid CI error messages?

^^ @ahelwer

muenchnerkindl:
So I removed the TLAPS module (from the repo and from manifest.json) and everything seems to work. Unless @ahelwer sees a reason for including the module, I think this is good to merge.

ahelwer:
@lemmy I recall I had deleted all the TLAPS.tla files in the big TLAPS CI PR but you wanted those changes reverted, don't remember why.

PR #113, opened 2024-01-22
Find solution to Einstein's riddle with Apalache

nano-o:
As you can see it does not look as good as before. The string constants now all have the _OF_TYPE prefix. I'm not sure how to do better. @konnov any suggestion?

### DIFF
--- a/specifications/EinsteinRiddle/Einstein.tla
+++ b/specifications/EinsteinRiddle/Einstein.tla
@@ -129,6 +129,8 @@ Init ==
     /\ pets \in PETS
     /\ cigars \in CIGARS
 
+\* Apalache cannot infer the type of `vars' because it could be a sequence or a tuple.
+\* So we explicitely tell Apalache that it is a sequence by adding the following annotation:
 \* @type: Seq(Int -> Str);
 vars == <<nationality, colors, cigars, pets, drinks>>
 
@@ -157,7 +159,7 @@ Solution ==
 FindSolution == ~Solution
 
 \* To find the solution with the `^Apalache^' model-checker, run:
-\* apalache-mc check --init=Init --inv=FindSolution --length=0 Einstein.tla
-\* Then look for the file violation.tla in `^Apalache^' output directory.
+\* `^apalache-mc check --init=Init --inv=FindSolution --length=0 --run-dir=./outout Einstein.tla^'
+\* You will then find the solution in `^./output/violation.tla^'.
 
 ============================================================
--- a/specifications/EinsteinRiddle/Einstein.tla
+++ b/specifications/EinsteinRiddle/Einstein.tla
@@ -35,12 +35,14 @@
 
 EXTENDS Naturals, FiniteSets
 
+House == 1..5
+
 \* Note that TLC!Permutations has a Java module override and, thus,
 \* would be evaluated faster.  However, TLC!Permutations equals a
 \* set of records whereas Permutation below equals a set of tuples/
 \* sequences.  Also, Permutation expects Cardinality(S) = 5.
 Permutation(S) == 
-    { p \in [ 1..5 -> S ] :
+    { p \in [ House -> S ] :
         /\ p[2] \in S \ {p[1]}
         /\ p[3] \in S \ {p[1], p[2]}
         /\ p[4] \in S \ {p[1], p[2], p[3]}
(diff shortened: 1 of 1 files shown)


nano-o:
I tried annotating all variables with the type `Int -> Str`, but then Apalache fails on the `Spec == ...` line.

specifications/EinsteinRiddle/Einstein.tla
```diff
@@ -51,17 +53,22 @@ Permutation(S) ==
 \* parameterization, the constants are replaced with constant-level
 \* operators.  As a side-effect, TLC evaluates them eagerly at startup, 
 \* and Apalache successfully determines their types.
-NATIONALITIES == Permutation({ "brit", "swede", "dane", "norwegian", "german" })
-DRINKS == Permutation({ "beer", "coffee", "mylk", "tea", "water" })
-COLORS == Permutation({ "red", "white", "blue", "green", "yellow" })
-PETS == Permutation({ "bird", "cat", "dog", "fish", "horse" })
-CIGARS == Permutation({ "blend", "bm", "dh", "pm", "prince" })
+NATIONALITIES == Permutation({ "brit_OF_NATIONALITY", "swede_OF_NATIONALITY", "dane_OF_NATIONALITY", "norwegian_OF_NATIONALITY", "german_OF_NATIONALITY" })
```
konnov, line 56:
```suggestion
BRIT == "brit_OF_NATIONALITY"
SWEDE == "swede_OF_NATIONALITY"
DANE == "dane_OF_NATIONALITY"
NORWEGIAN == "norwegian_OF_NATIONALITY"
GERMAN == "german_OF_NATIONALITY"
NATIONALITIES == Permutation({ BRIT, SWEDE, DANE, NORWEGIAN, GERMAN })
```

Here is my suggestion on how to avoid specifying the type all the time.

konnov:
Yeah, this look like a lot of literals that are just screaming their types all the time. What you could is to define a value for each literal and refer to that value. It will increase the size of the spec, but that's the easiest thing that comes to my mind:

https://github.com/tlaplus/Examples/pull/113/files#r1461491804

specifications/EinsteinRiddle/Einstein.tla
```diff
@@ -35,12 +35,14 @@
 
 EXTENDS Naturals, FiniteSets
 
+House == 1..5
+
 \* Note that TLC!Permutations has a Java module override and, thus,
 \* would be evaluated faster.  However, TLC!Permutations equals a
 \* set of records whereas Permutation below equals a set of tuples/
 \* sequences.  Also, Permutation expects Cardinality(S) = 5.
 Permutation(S) ==
```
konnov, line 44: Just by looking at the set of constraints, I think you could define `Permutation` for the general case too:

```tla
Permutation(S) ==
  { p \in [ House -> S ]:
    \A h1 \in House:
      \A h2 \in House:
        h2 /= h1 => p[h2] /= p[h1]
  }
```

You could even replace `House` with `S`, and it should still work in Apalache, as long as `S` is a finite set. Of course, this definition would produce really a lot of constraints, something like $O(|S|^{|House|}*|House|^2)$

konnov (approved):
This looks like an interesting example for TLA+ and Apalache.

nano-o:
> I tried annotating all variables with the type `Int -> Str`, but then Apalache fails on the `Spec == ...` line.

@konnov is this a bug I should report?

nano-o:
Now using`Str`. This allows to use the same string constants as in the original version.

konnov:
> > I tried annotating all variables with the type `Int -> Str`, but then Apalache fails on the `Spec == ...` line.
> 
> @konnov is this a bug I should report?

Yes, perhaps, something was missing in the annotations.

nano-o:
@lemmy is it acceptable to add type annotations to this spec or should they go into a separate file `ApaEinstein.tla`?

lemmy:
The specification might be easier to read if type annotations are placed in a separate file. Alternatively, enhanced IDE tools and pretty-printers could be used to minimize the visual impact of type annotations in comments, thereby preserving readability.

nano-o:
I'm not sure it will not be possible to move the annotation on `vars` to a separate file.

ahelwer:
I don't think this spec is useful without being checked by Apalache (since TLC just times out on the initial state iirc) so it seems okay to me to modify the spec itself.

specifications/EinsteinRiddle/Einstein.tla
```diff
@@ -146,4 +156,8 @@ Solution ==
 
 FindSolution == ~Solution
 
+\* To find the solution with the `^Apalache^' model-checker, run:
+\* apalache-mc check --init=Init --inv=FindSolution --length=0 Einstein.tla
+\* Then look for the file violation.tla in `^Apalache^' output directory.
```
shonfeder, line 161: In case it is helpful, you can use the `--run-dir` flag to put the output (including the violation) in a known location:

```
 apalache-mc --help | grep 'run-dir'
   --run-dir            : additional directory wherein output files for this run will be written directly, default: none (overrides envvar RUN_DIR)
```

ahelwer:
Is this ready to be evaluated & merged or is it a draft pull request?

nano-o:
> Is this ready to be evaluated & merged or is it a draft pull request?

Yes this is ready to be evaluated and merged.

PR #116, opened 2024-01-31
Move some specs from remote spec table into local repo

ahelwer:
Also update remote spec table format.

Specs added:

- Murat's PlusCal version of Lamport & Gray's analysis of Paxos vs. Two-Phase Commit
- Voucher Transaction System
- Atomic Commitment Protocol
- Byzantine Paxos


.ciignore
```diff
@@ -5,3 +5,8 @@
 # Example:
 specifications/doesnotexist
 
+# Ignore submodules to ensure deterministic local execution
```
lemmy, line 8: Can you please explain why they would introduce non-determinism?  Don't the submodule points to a particular commit?

ahelwer: Basically if some people have the repo checked out with submodules they'll generate a different manifest.json than if other people have the repo checked out without submodules.

lemmy: Is that a problem given that the authoritative manifest is generated by our GitHub action?

ahelwer: It isn't actually generated by the action, the action just validates that it's correct.

lemmy: Sounds good enough to catch it when somebody accidentally generates the manifest without the submodules m

ahelwer: I suppose it is, but putting these dirs in the .ciignore file saves whoever is trying to submit specs to the examples repo a CI round trip. It's easy to remove these if we want to add these dirs to the CI eventually, although at this point we literally do want to ignore them.

ahelwer: Ah you're talking about adding the submodule specs to the CI. Well, maybe. I will look into doing that after this work.

PR #121, opened 2024-02-20
Add Practical SMR-system-style MultiPaxos Spec

josehu07:
Adds a new MultiPaxos spec at `specifications/MultiPaxos-SMR`. This is a spec that very closely models how a practical state machine replication (SMR) system would implement MultiPaxos.

A more detailed description can be found in the added README in the spec folder.

Proper updates to both the repo README and the `manifest.json` have been made. I hope the CI runs as expected.


ahelwer:
Thanks!

PR #125, opened 2024-03-12
PlusCal variant of EWD998.

lemmy:
@muenchnerkindl Vanilla PlusCal variant of EWD998.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ It serves as:
 - a collection of case studies in the application of formal specification in TLA<sup>+</sup>
 
 All TLA<sup>+</sup> specs can be found in the [`specifications`](specifications) directory.
-The table below lists all specs and their features, such as whether they are suitable for beginners or use PlusCal.
+The table below lists all specs and their features, such as whether they are suitable for beginners, come with an additional PlusCal variant `(✔)`, or use PlusCal exclusively.
 The [`manifest.json`](manifest.json) file is the source of truth for this table, and is a detailed human- & machine-readable description of the specs & their models.
 Its schema is [`manifest-schema.json`](manifest-schema.json).
 All specs in this repository are subject to CI validation to ensure quality.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ Here is a list of specs included in this repository, with links to the relevant
 | [Misra Reachability Algorithm](specifications/MisraReachability)                                    | Leslie Lamport                                      |          |      ✔      |    ✔    |     ✔     |          |
 | [Byzantizing Paxos by Refinement](specifications/byzpaxos)                                          | Leslie Lamport                                      |          |      ✔      |    ✔    |     ✔     |          |
 | [EWD840: Termination Detection in a Ring](specifications/ewd840)                                    | Stephan Merz                                        |          |      ✔      |         |     ✔     |          |
-| [EWD998: Termination Detection in a Ring with Asynchronous Message Delivery](specifications/ewd998) | Stephan Merz, Markus Kuppe                          |          |      ✔      |         |     ✔     |          |
+| [EWD998: Termination Detection in a Ring with Asynchronous Message Delivery](specifications/ewd998) | Stephan Merz, Markus Kuppe                          |          |      ✔      |   (✔)   |     ✔     |          |
```
ahelwer, line 41: I don't know how the parentheses thing started, must have missed it - could you remove them and remove the parentheses from the other one that has them?

lemmy: I don't see a better way to indicate that a primary TLA+ spec also comes with an alternative Pluscal spec. Without parents, readers might think it is only a Pluscal spec.

ahelwer: Perhaps then also add a note about the parentheses meaning above the table.

lemmy: https://github.com/tlaplus/Examples/pull/125/commits/9bed7d54bc73e0f1cce978a34491f8825ea0cb59#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R12

manifest.json
```diff
@@ -4330,6 +4330,28 @@
...
+          "path": "specifications/ewd998/EWD998PCal.tla",
+          "communityDependencies": [
+            "BagsExt"
+          ],
+          "tlaLanguageVersion": 2,
+          "features": ["pluscal"],
+          "models": [
+            {
+              "path": "specifications/ewd998/EWD998PCal.cfg",
+              "runtime": "00:01:00",
```
ahelwer, line 4343: If the runtime is one minute this spec doesn't count as a small spec; those are for specs that execute in less than 10 seconds.

PR #129, opened 2024-03-27
Fix incorrect tar flag in install.sh of devcontainer

melhindi:
The `z` flag in the `tar` command in the `install.sh` script was on the wrong position. This caused the following error: `tar: z: Cannot open: No such file or directory`.

Making `f` the last flag before the filename and moving the `z` to the front fixes the issue.


ahelwer:
You'll need to sign off with a DCO, read https://github.com/tlaplus/Examples/pull/129/checks?check_run_id=23145024735

lemmy:
Where/What OS does this cause issues?

PR #136, opened 2024-04-10
Test PlusCal translation in CI

ahelwer:
The CI will fail if there is invalid PlusCal syntax in any of the modules. Previously, SANY would only warn of this but would return a zero exit code (see https://github.com/tlaplus/tlaplus/issues/899).

Also update the PlusCal translations of a number of modules with trivial differences.

### DIFF
--- a/specifications/DiningPhilosophers/DiningPhilosophers.tla
+++ b/specifications/DiningPhilosophers/DiningPhilosophers.tla
@@ -113,7 +113,7 @@ Think:
 end process;
 
 end algorithm; *)
-\* BEGIN TRANSLATION (chksum(pcal) = "8d8268d4" /\ chksum(tla) = "16352822")
+\* BEGIN TRANSLATION (chksum(pcal) = "ea877089" /\ chksum(tla) = "16352822")
 VARIABLES forks, pc
 
 (* define statement *)
--- a/.github/scripts/translate_pluscal.py
+++ b/.github/scripts/translate_pluscal.py
@@ -1,5 +1,5 @@
 """
-Converts all TLA⁺ modules from ASCII to Unicode or vice-versa.
+Runs PlusCal translation on all PlusCal specs.
 """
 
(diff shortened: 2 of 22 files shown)


specifications/DiningPhilosophers/DiningPhilosophers.tla
```diff
@@ -113,7 +113,7 @@ Think:
 end process;
 
 end algorithm; *)
-\* BEGIN TRANSLATION (chksum(pcal) = "8d8268d4" /\ chksum(tla) = "16352822")
+\* BEGIN TRANSLATION (chksum(pcal) = "ea877089" /\ chksum(tla) = "16352822")
```
lemmy, line 116: FYI, the Toolbox raises a warning if the PlusCal or the translated TLA+ do not match their checksums, respectively. However, a user may write `chksum(pcal) \in STRING` or `chksum(tla) \in STRING` to indicate that the pcal or TLA+ has been edited after translation. If the goal of this PR is to always retranslate, I'd suggest honoring the same convention and skipping retranslation.

ahelwer: Hmmm okay, so this would be as simple as searching for the string `chksum(pcal) \in STRING` in the file then skipping the translation if so?

lemmy: The logic is implemented in the Pcal translator: https://github.com/tlaplus/tlaplus/blob/master/tlatools/org.lamport.tlatools/src/pcal/Validator.java

ahelwer: Okay so it's fine to re-run translation on every file? Also thinking about emitting an error in the CI if running the translator results in a mismatch.

lemmy: Sounds fine to me to report an error if there is a checksum mismatch.

PR #137, opened 2024-04-14
Remove pluscal checksums

ahelwer:
Ensure `pcal.trans` always re-translates the files.

### DIFF
--- a/specifications/DiningPhilosophers/DiningPhilosophers.tla
+++ b/specifications/DiningPhilosophers/DiningPhilosophers.tla
@@ -113,7 +113,7 @@ Think:
 end process;
 
 end algorithm; *)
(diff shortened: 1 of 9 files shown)


lemmy (changes requested):
The checksums are perfectly fine and help users if they work with the specs locally.  I don't think they should be removed.

ahelwer:
Hmmm, perhaps I can just remove them via a regex in the CI before running the pluscal translator. Should the checksums be added to the pluscal modules that don't have them?

lemmy:
Perhaps, the pcal translator needs a "force flag" (system property) to retranslate even if checksums match.  That way, you won't have to modify the .tla files in the CI.

PR #141, opened 2024-04-20
Add Apalache to CI

ahelwer:
This adds Apalache to the CI along with support for using it to check models marked as using the symbolic model-checking mode. Since this requires an extra parameter to be added to a number of scripts, this is a breaking change.

@konnov as FYI


lemmy:
I've alreay merged tlaplus/tlaplus#912 as this seems ready.

PR #143, opened 2024-05-13
Add DOI of paper where ClientCentric TLA+ specifications are documented

TimSoethout:
Sync up with https://github.com/cwi-swat/tla-ci/blob/master/ClientCentric.tla


muenchnerkindl:
Thanks for adding this pointer.

PR #145, opened 2024-06-14
Add BufferedRandomAccessFile spec

Calvin-L:
In review, @lemmy suggested adding this spec to our examples: https://github.com/tlaplus/tlaplus/pull/907#discussion_r1569827552

I made two small changes:
 1. Pull config out into a separate file
 2. Reduce `MaxOffset` from 4 to 3 so it can be checked in <10s (er, on my 8-core laptop anyway)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -3416,6 +3416,38 @@
       ]
     },
     {
+      "path": "specifications/braf",
+      "title": "Buffered Random Access File",
+      "description": "Specification of a Java file caching layer",
+      "sources": ["https://github.com/tlaplus/tlaplus/tree/4613109641676389d97b8df209d6cf4d90d31c1c/tlatools/org.lamport.tlatools/src/tlc2/util/BufferedRandomAccessFile.tla"],
+      "authors": ["Calvin Loncaric"],
+      "tags": [],
+      "modules": [
+        {
+          "path": "specifications/braf/BufferedRandomAccessFile.tla",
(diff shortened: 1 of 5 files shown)


lemmy:
Duplication of the specification, both here and in the tlaplus/tlaplus repository, likely won't be an issue. However, it's important to note that the OpenAddressing.tla specification in the tlaplus/tlaplus repository is only referenced here, not duplicated.

ahelwer:
This is very weird. CI has SANY output the following error:
```
ERROR:root:java -cp deps/tools/tla2tools.jar:deps/apalache/lib/apalache.jar:specifications/Bakery-Boulangerie:deps/community/modules.jar:deps/tlapm/library tla2sany.SANY -error-codes specifications/Bakery-Boulangerie/MCBakery.tla
Illegal switch: -error-codes
```

manifest.json
```diff
@@ -3415,6 +3415,39 @@
...
+      "modules": [
+        {
+          "path": "specifications/braf/BufferedRandomAccessFile.tla",
+          "communityDependencies": [],
+          "tlaLanguageVersion": 2,
+          "features": [],
+          "models": [
+            {
+              "path": "specifications/braf/BufferedRandomAccessFile.cfg",
+              "runtime": "00:09:00",
```
ahelwer, line 3434: Time format is HH:MM:SS

Issue #119, opened 2024-02-08
Deactivate macOS CI runners?


ahelwer:
Currently the macOS CI takes over twice as long to run as the linux and windows CI runs: https://github.com/tlaplus/Examples/actions/runs/7835386801

This got substantially worse after the work to add models to most of the specs, because macOS CI runners see a strange amount of overhead for each model (for example a model taking 1 second to run on the other runners takes 8 seconds to run on the macOS runner). Looking online it seems slow macOS CI runners is a common problem experienced by developers across github.

Ideally the CI would run for ten minutes at most, and I don't think the macOS runner is providing enough value (given the similarity to linux) to justify keeping it around. Perhaps we could add a scheduled CI run to test on macOS once per week.

lemmy:
AFAIK there are also much faster m1-based Github runners now.

ahelwer:
Ah, you have to specify macos-14 instead of macos-latest to get those. Running the CI with that it does seem as though it's faster. Great, problem solved!

lemmy:
I recommend reactivating macOS for the same reason it was kept for [tlaplus/tlaplus](https://github.com/tlaplus/tlaplus/issues/986). macOS is the most popular platform and shares similarities with Linux, though it is not identical. Additionally, the rate of changes in this repository is very low, so allowing 30 minutes instead of 15 minutes for tasks is acceptable.

PR #149, opened 2024-07-18
Use macOS runner, minor script debug output fixes

ahelwer:
Print out TLC command line call in smoke testing script and decode TLC output as UTF-8 if necessary.
Ref #119

### DIFF
--- a/.github/scripts/smoke_test_large_models.py
+++ b/.github/scripts/smoke_test_large_models.py
@@ -48,15 +48,25 @@ def check_model(module_path, model):
     )
     match tlc_result:
         case TimeoutExpired():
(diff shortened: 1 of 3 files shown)


lemmy:
What does the minor script fix do?

ahelwer:
Sometimes (varies depending on OS and also way in which the process terminated) stdout is given in bytes, and sometimes it's given as a string. So this just ensures it's always printed out as a string in the smoke testing script.

lemmy:
I'd suggest to make that the git commit message for that particular change.

ahelwer:
Sure, I will put that in the squash commit message.

PR #150, opened 2024-09-13
Add spec of the Disruptor concurrency library.

nicholassm:
The `Disruptor` is a concurrency library originally developed and open sourced by LMAX Exchange for low latency communication via a ring buffer between producer and consumer threads.

This PR adds a spec of the Disruptor lib and verifies that data races do not occur.

### DIFF
--- a/specifications/Disruptor/Disruptor_SPMC.cfg
+++ b/specifications/Disruptor/Disruptor_SPMC.cfg
@@ -6,7 +6,7 @@ CONSTANTS
     Readers = { r1 r2 r3 }
     MaxPublished = 10
     Size = 4
-    NULL = [ model value ]
+    NULL = NULL
 
 INVARIANTS
     TypeOK
--- a/specifications/Disruptor/Disruptor_SPMC.cfg
+++ b/specifications/Disruptor/Disruptor_SPMC.cfg
@@ -2,9 +2,9 @@ SPECIFICATION
     Spec
 
 CONSTANTS
+    MaxPublished = 10
     Writers = { w }
     Readers = { r1, r2, r3 }
-    MaxPublished = 10
     Size = 4
     NULL = NULL
 
--- a/specifications/Disruptor/Disruptor_SPMC.cfg
+++ b/specifications/Disruptor/Disruptor_SPMC.cfg
@@ -1,0 +1,16 @@
+SPECIFICATION
+    Spec
+
+CONSTANTS
+    Writers = { w }
+    Readers = { r1 r2 r3 }
+    MaxPublished = 10
+    Size = 4
+    NULL = [ model value ]
+
+INVARIANTS
+    TypeOK
+    NoDataRaces
+
+PROPERTIES
+    Liveliness
--- a/specifications/Disruptor/Disruptor_SPMC.cfg
+++ b/specifications/Disruptor/Disruptor_SPMC.cfg
@@ -3,14 +3,17 @@ SPECIFICATION
 
 CONSTANTS
     Writers = { w }
-    Readers = { r1 r2 r3 }
+    Readers = { r1, r2, r3 }
     MaxPublished = 10
     Size = 4
     NULL = NULL
 
 INVARIANTS
-    TypeOK
+    TypeOk
     NoDataRaces
 
 PROPERTIES
     Liveliness
+
+CHECK_DEADLOCK
+    FALSE
--- a/specifications/Disruptor/Disruptor_SPMC.tla
+++ b/specifications/Disruptor/Disruptor_SPMC.tla
@@ -22,6 +22,8 @@ CONSTANTS
   Size,         (* Ringbuffer size.                                  *)
   NULL
 
+ASSUME Writers /= {}
+ASSUME Readers /= {}
 ASSUME Size \in Nat \ {0}
 
 VARIABLES
--- a/specifications/Disruptor/Disruptor_SPMC.tla
+++ b/specifications/Disruptor/Disruptor_SPMC.tla
@@ -1,0 +1,155 @@
+----------------------------- MODULE Disruptor_SPMC -------------------------
+(***************************************************************************)
+(* Models a Single Producer, Multi Consumer Disruptor (SPMC).              *)
+(*                                                                         *)
+(* The model verifies that no data races occur between the publisher       *)
+(* and consumers and that all consumers eventually read all published      *)
+(* values.                                                                 *)
+(*                                                                         *)
+(* To see a data race, try and run the model with two publishers.          *)
+(***************************************************************************)
+
+EXTENDS Naturals, Integers, FiniteSets, Sequences
+
+CONSTANTS
+  Writers,      (* Writer/publisher thread ids.    *)
+  Readers,      (* Reader/consumer  thread ids.    *)
+  MaxPublished, (* Max number of published events. *)
+  Size,         (* Ringbuffer size.                *)
+  NULL
+
+VARIABLES
+  ringbuffer,
+  published,    (* Publisher Cursor.                           *)
+  read,         (* Read Cursors. One per consumer.             *)
+  consumed,     (* Sequence of all read events by the Readers. *)
+  pc            (* Program Counter of each Writer/Reader.      *)
+
+vars == <<
+  ringbuffer,
+  published,
+  read,
+  consumed,
+  pc
+>>
+
+(***************************************************************************)
+(* Each publisher/consumer can be in one of two states:                    *)
+(* 1. Accessing a slot in the Disruptor or                                 *)
+(* 2. Advancing to the next slot.                                          *)
+(***************************************************************************)
+Access  == "Access"
+Advance == "Advance"
+
+Transition(t, from, to) ==
+  /\ pc[t] = from
+  /\ pc'   = [ pc EXCEPT ![t] = to ]
+
+Buffer  == INSTANCE RingBuffer
+
+Range(f) ==
+  { f[x] : x \in DOMAIN(f) }
+
+MinReadSequence ==
+  CHOOSE min \in Range(read) : \A r \in Readers : min <= read[r]
+
+(***************************************************************************)
+(* Publisher Actions:                                                      *)
+(***************************************************************************)
+
+BeginWrite(writer) ==
+  LET
+    next     == published + 1
+    index    == Buffer!IndexOf(next)
+    min_read == MinReadSequence
+  IN
+    \* Are we clear of all consumers? (Potentially a full cycle behind).
+    /\ min_read >= next - Size
+    /\ next < MaxPublished
+    /\ Transition(writer, Access, Advance)
+    /\ Buffer!Write(index, writer, next)
+    /\ UNCHANGED << consumed, published, read >>
+
+EndWrite(writer) ==
+  LET
+    next  == published + 1
+    index == Buffer!IndexOf(next)
+  IN
+    /\ Transition(writer, Advance, Access)
+    /\ Buffer!EndWrite(index, writer)
+    /\ published' = next
+    /\ UNCHANGED << consumed, read >>
+
+(***************************************************************************)
+(* Consumer Actions:                                                       *)
+(***************************************************************************)
+
+BeginRead(reader) ==
+  LET
+    next  == read[reader] + 1
+    index == Buffer!IndexOf(next)
+  IN
+    /\ published >= next
+    /\ Transition(reader, Access, Advance)
+    /\ Buffer!BeginRead(index, reader)
+    \* Track what we read from the ringbuffer.
+    /\ consumed' = [ consumed EXCEPT ![reader] = Append(@, Buffer!Read(index)) ]
+    /\ UNCHANGED << published, read >>
+
+EndRead(reader) ==
+  LET
+    next  == read[reader] + 1
+    index == Buffer!IndexOf(next)
+  IN
+    /\ Transition(reader, Advance, Access)
+    /\ Buffer!EndRead(index, reader)
+    /\ read' = [ read EXCEPT ![reader] = next ]
+    /\ UNCHANGED << consumed, published >>
+
+(***************************************************************************)
+(* Spec:                                                                   *)
+(***************************************************************************)
+
+Init ==
+  /\ Buffer!Init
+  /\ published = -1
+  /\ read      = [ r \in Readers                |-> -1     ]
+  /\ consumed  = [ r \in Readers                |-> << >>  ]
+  /\ pc        = [ a \in Writers \union Readers |-> Access ]
+
+Next ==
+  \/ \E w \in Writers : BeginWrite(w)
+  \/ \E w \in Writers : EndWrite(w)
+  \/ \E r \in Readers : BeginRead(r)
+  \/ \E r \in Readers : EndRead(r)
+
+Fairness ==
+  /\ \A w \in Writers : WF_vars(BeginWrite(w))
+  /\ \A w \in Writers : WF_vars(EndWrite(w))
+  /\ \A r \in Readers : WF_vars(BeginRead(r))
+  /\ \A r \in Readers : WF_vars(EndRead(r))
+
+Spec ==
+  Init /\ [][Next]_vars /\ Fairness
+
+(***************************************************************************)
+(* Invariants:                                                             *)
+(***************************************************************************)
+
+TypeOk ==
+  /\ Buffer!TypeOk
+  /\ published \in Int
+  /\ read      \in [ Readers                -> Int                 ]
+  /\ consumed  \in [ Readers                -> Seq(Nat)            ]
+  /\ pc        \in [ Writers \union Readers -> { Access, Advance } ]
+
+NoDataRaces == Buffer!NoDataRaces
+
+(***************************************************************************)
+(* Properties:                                                             *)
+(***************************************************************************)
+
+Liveliness ==
+  <>[] (\A r \in Readers : consumed[r] = [i \in 1..MaxPublished |-> i - 1])
+
+=============================================================================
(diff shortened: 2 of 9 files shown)


ahelwer:
Nice spec! Thanks for the contribution.

nicholassm:
Thanks, you're welcome. :-)

nicholassm:
I can see a lot of checks failing due to different number of states, distinct states, etc. compared to what I get when running the toolbox. Can it be a TLC version issue? Or symmetry vs. non-symmetry sets for model values that differ somehow?

Kind regards
Nicholas

ahelwer:
Those json fields are optional, although useful as a regression test. I will take a look tomorrow.

nicholassm:
It looks like it's not my model that fails but a module called `MCtcp` that results in an exit code 1... What gives?

specifications/Disruptor/Disruptor_SPMC.tla
```diff
@@ -0,0 +1,155 @@
...
+(*                                                                         *)
+(* To see a data race, try and run the model with two publishers.          *)
+(***************************************************************************)
+
+EXTENDS Naturals, Integers, FiniteSets, Sequences
+
+CONSTANTS
+  Writers,      (* Writer/publisher thread ids.    *)
+  Readers,      (* Reader/consumer  thread ids.    *)
+  MaxPublished, (* Max number of published events. *)
```
muenchnerkindl, line 17: I presume this constant is only relevant for model checking? It would be cleaner to separate the logical spec from the bounds imposed for model checking and either add a state constraint such as ```published < MaxPublished``` in the cfg file or write a MC version of the spec that adds extra guards to actions.

specifications/Disruptor/Disruptor_SPMC.cfg
```diff
@@ -0,0 +1,16 @@
+SPECIFICATION
+    Spec
+
+CONSTANTS
+    Writers = { w }
+    Readers = { r1 r2 r3 }
```
muenchnerkindl, line 6: This should be ```{r1, r2, r3}```.

specifications/Disruptor/Disruptor_SPMC.cfg
```diff
@@ -0,0 +1,16 @@
...
+    MaxPublished = 10
+    Size = 4
+    NULL = NULL
+
+INVARIANTS
+    TypeOK
```
muenchnerkindl, line 12: written ```TypeOk``` in the TLA module

specifications/Disruptor/Disruptor_SPMC.cfg
```diff
@@ -0,0 +1,16 @@
...
+    NoDataRaces
+
+PROPERTIES
+    Liveliness
```
muenchnerkindl, line 16: You probably want to add ```CHECK_DEADLOCK FALSE``` so that model checking completes.

specifications/Disruptor/Disruptor_SPMC.tla
```diff
@@ -0,0 +1,155 @@
...
+  Size,         (* Ringbuffer size.                *)
+  NULL
+
```
muenchnerkindl, line 20: Probably want to add some assumptions such as ```Size \in Nat \ {0}" or ```NULL \notin Int```.

ahelwer:
Sorry, the spec checking has debug output enabled so it's very verbose. You want to search for the text `ERROR:root` in the raw log. @lemmy should we keep debug logging enabled for model-checking in the CI? It makes it difficult to sift through the output.

Anyway here is the result:
```
2024-09-13T21:10:08.0204980Z INFO:root:specifications/Disruptor/Disruptor_MPMC.cfg
2024-09-13T21:10:08.2670270Z INFO:root:specifications/Disruptor/Disruptor_MPMC.cfg in 0.3s vs. 10s expected
2024-09-13T21:10:08.2671430Z ERROR:root:Model specifications/Disruptor/Disruptor_MPMC.cfg expected result success but got 255
2024-09-13T21:10:08.2674970Z ERROR:root:java -enableassertions -Dtlc2.TLC.ide=Github -Dutil.ExecutionStatisticsCollector.id=abcdef60f238424fa70d124d0c77ffff -XX:+UseParallelGC -cp deps/tools/tla2tools.jar:deps/apalache/bin/apalache-mc/lib/apalache.jar:deps/community/modules.jar:deps/tlapm/library tlc2.TLC specifications/Disruptor/Disruptor_MPMC.tla -config specifications/Disruptor/Disruptor_MPMC.cfg -workers auto -lncheck final -cleanup
2024-09-13T21:10:08.2677880Z TLC2 Version 2.20 of Day Month 20?? (rev: caf0c33)
2024-09-13T21:10:08.2680380Z Running breadth-first search Model-Checking with fp 31 and seed -471696565524114047 with 3 workers on 3 cores with 1593MB heap and 64MB offheap memory [pid: 5082] (Mac OS X 14.6.1 aarch64, Eclipse Adoptium 17.0.12 x86_64, MSBDiskFPSet, DiskStateQueue).
2024-09-13T21:10:08.2682070Z Error: TLC threw an unexpected exception.
2024-09-13T21:10:08.2682650Z This was probably caused by an error in the spec or model.
2024-09-13T21:10:08.2683300Z See the User Output or TLC Console for clues to what happened.
2024-09-13T21:10:08.2683890Z The exception was a tlc2.tool.ConfigFileException
2024-09-13T21:10:08.2684460Z : TLC found an error in the configuration file at line 5
2024-09-13T21:10:08.2685000Z It was expecting }, but did not find it.
2024-09-13T21:10:08.2685510Z Finished in 00s at (2024-09-13 21:10:08)
2024-09-13T21:10:08.2685800Z 
2024-09-13T21:10:08.2686070Z INFO:root:specifications/Disruptor/Disruptor_MPMC_liveliness.cfg
2024-09-13T21:10:08.5685730Z INFO:root:specifications/Disruptor/Disruptor_MPMC_liveliness.cfg in 0.3s vs. 10s expected
2024-09-13T21:10:08.5687240Z ERROR:root:Model specifications/Disruptor/Disruptor_MPMC_liveliness.cfg expected result success but got 255
2024-09-13T21:10:08.5691920Z ERROR:root:java -enableassertions -Dtlc2.TLC.ide=Github -Dutil.ExecutionStatisticsCollector.id=abcdef60f238424fa70d124d0c77ffff -XX:+UseParallelGC -cp deps/tools/tla2tools.jar:deps/apalache/bin/apalache-mc/lib/apalache.jar:deps/community/modules.jar:deps/tlapm/library tlc2.TLC specifications/Disruptor/Disruptor_MPMC.tla -config specifications/Disruptor/Disruptor_MPMC_liveliness.cfg -workers auto -lncheck final -cleanup
2024-09-13T21:10:08.5695900Z TLC2 Version 2.20 of Day Month 20?? (rev: caf0c33)
2024-09-13T21:10:08.5698330Z Running breadth-first search Model-Checking with fp 130 and seed -5831432232768886009 with 3 workers on 3 cores with 1593MB heap and 64MB offheap memory [pid: 5083] (Mac OS X 14.6.1 aarch64, Eclipse Adoptium 17.0.12 x86_64, MSBDiskFPSet, DiskStateQueue).
2024-09-13T21:10:08.5700840Z Error: TLC threw an unexpected exception.
2024-09-13T21:10:08.5701540Z This was probably caused by an error in the spec or model.
2024-09-13T21:10:08.5702440Z See the User Output or TLC Console for clues to what happened.
2024-09-13T21:10:08.5703240Z The exception was a tlc2.tool.ConfigFileException
2024-09-13T21:10:08.5704050Z : TLC found an error in the configuration file at line 5
2024-09-13T21:10:08.5704850Z It was expecting }, but did not find it.
2024-09-13T21:10:08.5705510Z Finished in 00s at (2024-09-13 21:10:08)
2024-09-13T21:10:08.5705920Z 
2024-09-13T21:10:08.5706210Z INFO:root:specifications/Disruptor/Disruptor_SPMC.cfg
2024-09-13T21:10:08.8707050Z INFO:root:specifications/Disruptor/Disruptor_SPMC.cfg in 0.3s vs. 10s expected
2024-09-13T21:10:08.8708400Z ERROR:root:Model specifications/Disruptor/Disruptor_SPMC.cfg expected result success but got 255
2024-09-13T21:10:08.8712860Z ERROR:root:java -enableassertions -Dtlc2.TLC.ide=Github -Dutil.ExecutionStatisticsCollector.id=abcdef60f238424fa70d124d0c77ffff -XX:+UseParallelGC -cp deps/tools/tla2tools.jar:deps/apalache/bin/apalache-mc/lib/apalache.jar:deps/community/modules.jar:deps/tlapm/library tlc2.TLC specifications/Disruptor/Disruptor_SPMC.tla -config specifications/Disruptor/Disruptor_SPMC.cfg -workers auto -lncheck final -cleanup
2024-09-13T21:10:08.8716690Z TLC2 Version 2.20 of Day Month 20?? (rev: caf0c33)
2024-09-13T21:10:08.8719090Z Running breadth-first search Model-Checking with fp 9 and seed 5823395496300575269 with 3 workers on 3 cores with 1593MB heap and 64MB offheap memory [pid: 5084] (Mac OS X 14.6.1 aarch64, Eclipse Adoptium 17.0.12 x86_64, MSBDiskFPSet, DiskStateQueue).
2024-09-13T21:10:08.8721110Z Error: TLC threw an unexpected exception.
2024-09-13T21:10:08.8721820Z This was probably caused by an error in the spec or model.
2024-09-13T21:10:08.8722650Z See the User Output or TLC Console for clues to what happened.
2024-09-13T21:10:08.8723490Z The exception was a tlc2.tool.ConfigFileException
2024-09-13T21:10:08.8724220Z : TLC found an error in the configuration file at line 6
2024-09-13T21:10:08.8724960Z It was expecting }, but did not find it.
2024-09-13T21:10:08.8726340Z Finished in 00s at (2024-09-13 21:10:08)
2024-09-13T21:10:08.8726770Z 
```

I believe the reason for the failure is that in your model files you have:
```
Writers = { w1 w2 }
```
but this needs to be comma-delimited like:
```
Writers = { w1, w2 }
```

muenchnerkindl (commented):
Thank you for the nice contribution. You'll find a few comments in the individual files (comments for the multi-writer version are analogous to those for the single-writer version). I am looking forward to seeing this spec added to the collection!

specifications/Disruptor/Disruptor_SPMC.cfg
ahelwer: If you do this the CI also wants you to add `"ignore deadlock"` to the model features in the `manifest.json`.

specifications/Disruptor/Disruptor_SPMC.cfg
nicholassm: Totally blunder - it was probably too late in the evening for me. :-P

nicholassm:
Hi @muenchnerkindl and @ahelwer - super nice with all the feedback - much appreciated.
I'll get busy incorporating your suggestions.

specifications/Disruptor/Disruptor_SPMC.tla
nicholassm: Yes and no. I've investigated and the constant does bound the model. So I agree - that would be better to have as a state constraint. But then I can't model check the liveliness property in the model (that all consumers eventually always read all published events) as liveliness cannot be verified when state or action constraints are specified.
Is there a "clean" third option that you know of?

nicholassm:
Hi guys, anything else I can do to improve the contribution? While it's fresh in my head :-)
Many thanks.

specifications/Disruptor/Disruptor_SPMC.tla
muenchnerkindl: Apologies for the late reply: now I am confused. First, I would rather have defined the liveness property as
```
Liveliness == \A r \in Readers : \A i \in 1 .. MaxPublished :
  <>[](i \in 1 .. published => Len(consumed[r]) >= i /\ consumed[r][i] = i-1)
```
(with the second conjunct on the right-hand side being optional). The reason is that the property that you assert will not hold if writers are allowed to continue publishing even after reaching the (artificial) bound since readers would then be able to update their ```consumed``` sequence as well, whereas the above property should always hold. (Even more general, the bound on ```i``` could be ```Nat``` but that would require an override of ```Nat``` during model checking so that TLC doesn't complain about an infinite quantifier bound, and that would be a little heavy.)

I then tried commenting out the guard ```next < MaxPublished``` in ```BeginWrite``` and adding the constraint
```
StateConstraint == published <= MaxPublished
```
for model checking. This appears to work just fine, and when I comment out the fairness condition on ```BeginRead```, TLC gives me the expected failure of the temporal property.

I would even suggest removing the fairness condition on ```BeginWrite``` from the spec, since why should writers be required to publish items forever? The above liveness property (but of course not the original one) continues to hold for the modified spec.

Of course, it's up to you to decide what you intend as your specification.

nicholassm: Thanks for the detailed description. I'll dive deeper into it in the weekend when I have more time. I'm keen on improving the spec (and learn!) :-)

specifications/Disruptor/Disruptor_SPMC.tla
```diff
@@ -0,0 +1,164 @@
...
+ASSUME Writers /= {}
+ASSUME Readers /= {}
+ASSUME Size \in Nat \ {0}
+
+VARIABLES
+  ringbuffer,
+  published,    (* Write cursor. One for the producer.               *)
+  read,         (* Read cursors. One per consumer.                   *)
+  consumed,     (* Sequence of all read events by the Readers.       *)
```
lemmy, line 33: `consumed` is a history variable. It might be helpful to add a comment explaining its purpose to make it clearer for readers.

specifications/Disruptor/Disruptor_SPMC.tla
lemmy: Stephan's `Liveliness` property would allow to model `consumed` as a counter instead of a sequence (for each reader). This observation also led me to realize that the SPMC specification seems to describe the [multicast configuration](https://lmax-exchange.github.io/disruptor/disruptor.htm), since all readers are consuming every value. It might be helpful to include this in a comment for clarity.

specifications/Disruptor/Disruptor_SPMC.tla
```diff
@@ -0,0 +1,164 @@
...
+  pc            (* Program Counter of each Writer/Reader.            *)
```
lemmy, line 34: Nit: Since the value for each process only alternates between `Advance` and `Access`, I would consider renaming the variable to something more descriptive, like `hasAccess`.

specifications/Disruptor/Disruptor_SPMC.tla
lemmy: If `consumed` remains a sequence, you could add the following "action property" (with `IsPrefix` defined at https://github.com/tlaplus/CommunityModules/blob/master/modules/SequencesExt.tla#L219-L225 :

```tla
\* We only ever append to the history variable  consumed  .   
Increments ==
    [][\A r \in Readers: IsPrefix(consumed[r], consumed'[r])]_vars
```

nicholassm: > Stephan's `Liveliness` property would allow to model `consumed` as a counter instead of a sequence (for each reader). This observation also led me to realize that the SPMC specification seems to describe the [multicast configuration](https://lmax-exchange.github.io/disruptor/disruptor.htm), since all readers are consuming every value. It might be helpful to include this in a comment for clarity.

Hi @lemmy, I wrote in the top that it's a Single Producer Multiple Consumer Disruptor - that's where I thought I made it clear that it is *that* configuration of the Disruptor. But perhaps I am assuming that everyone knows all consumers read all events (i.e. "multicast" behaviour) and that is not common knowledge?

specifications/Disruptor/Disruptor_SPMC.tla
nicholassm: Thanks, I'll add more comments explaining that. It's a new term to me but it makes total sense and it's nice to point out in the spec that's it's not part of the system being modeled but rather an auxiliary variable used for verification.

specifications/Disruptor/Disruptor_SPMC.tla
nicholassm: > Apologies for the late reply: now I am confused. First, I would rather have defined the liveness property as
> 
> ```
> Liveliness == \A r \in Readers : \A i \in 1 .. MaxPublished :
>   <>[](i \in 1 .. published => Len(consumed[r]) >= i /\ consumed[r][i] = i-1)
> ```
> 
> (with the second conjunct on the right-hand side being optional). The reason is that the property that you assert will not hold if writers are allowed to continue publishing even after reaching the (artificial) bound since readers would then be able to update their `consumed` sequence as well, whereas the above property should always hold. (Even more general, the bound on `i` could be `Nat` but that would require an override of `Nat` during model checking so that TLC doesn't complain about an infinite quantifier bound, and that would be a little heavy.)
> 
> I then tried commenting out the guard `next < MaxPublished` in `BeginWrite` and adding the constraint
> 
> ```
> StateConstraint == published <= MaxPublished
> ```
> 
> for model checking. This appears to work just fine, and when I comment out the fairness condition on `BeginRead`, TLC gives me the expected failure of the temporal property.
> 
> I would even suggest removing the fairness condition on `BeginWrite` from the spec, since why should writers be required to publish items forever? The above liveness property (but of course not the original one) continues to hold for the modified spec.
> 
> Of course, it's up to you to decide what you intend as your specification.

I did the changes (see a new PR) and I quite like it: Getting rid of the artificial model constraint in the `BeginWrite` action, remove the requirement to have producers publish forever and moving the bounding of the model to a state constraint. Elegant and a better model.

I still have some work to do for the MPMC model as it's more complex to write the liveliness property because there's multiple producers and hence it more difficult to express what `sequence` number a consumer can actually read. (More details will follow).

muenchnerkindl: Glad to hear that it worked out! I didn't look very much at the MPMC model, please let me know if you want me to.

PR #151, opened 2024-09-22
Add small improvements to the Disruptor spec.

nicholassm:
- Add comment on 'Multicast' behaviour.
- Add ASSUME for MaxPublished constant.
- Add comment on the usage of the 'consumed' variable.
- Remove WF from write actions.
- Remove bounding of model happening in BeginWrite action. Use a State constraint instead. Use suggestion for liveliness property.
- (The last two items are not done yet for the MPMC spec - it's less straight forward to do because of the multiple producers behaviour.)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -494,6 +494,7 @@
               "size": "small",
               "mode": "exhaustive search",
               "features": [
+                "state constraint",
                 "liveness",
                 "ignore deadlock"
               ],
--- a/manifest.json
+++ b/manifest.json
@@ -459,11 +459,12 @@
               "size": "small",
               "mode": "exhaustive search",
               "features": [
+                "state constraint",
                 "ignore deadlock"
               ],
               "result": "success",
               "distinctStates": 112929,
-              "totalStates": 406505,
+              "totalStates": 422781,
               "stateDepth": 81
             },
             {
@@ -472,12 +473,13 @@
               "size": "small",
               "mode": "exhaustive search",
               "features": [
+                "state constraint",
                 "liveness",
                 "ignore deadlock"
               ],
               "result": "success",
               "distinctStates": 14365,
-              "totalStates": 42101,
+              "totalStates": 44581,
               "stateDepth": 61
             }
           ]
--- a/manifest.json
+++ b/manifest.json
@@ -498,9 +498,9 @@
                 "ignore deadlock"
               ],
               "result": "success",
-              "distinctStates": 8153,
-              "totalStates": 26481,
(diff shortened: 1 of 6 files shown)


specifications/Disruptor/Disruptor_SPMC.tla
```diff
@@ -132,14 +134,18 @@ Next ==
   \/ \E r \in Readers : EndRead(r)
 
 Fairness ==
-  /\ \A w \in Writers : WF_vars(BeginWrite(w))
-  /\ \A w \in Writers : WF_vars(EndWrite(w))
```
lemmy, line 136: I find it surprising that a started write doesn't have to be ended.  Isn't the `EndWrite` action necessary so that another writer can eventually claim that slot?

nicholassm: This is the single producer case so there's just one.
Also, I think it actually makes sense for a producer to stutter after it has begun a write. It just means the consumers will never read that event. But on the other hand, the consumers will read the event if a producer finishes a write (as described in the liveliness property.)

nicholassm:
Alright, added the same improvements to the Multi Producer Multi Consumer specification.
Thanks.

manifest.json
```diff
@@ -494,13 +496,14 @@
               "size": "small",
               "mode": "exhaustive search",
               "features": [
+                "state constraint",
```
ahelwer, line 499: State constraints and liveness generally do not play well with each other, was this taken into account?

nicholassm: Yes, and I verified that it will be broken if it's not fulfilled (by putting in a temporary error in the model).

nicholassm:
I'm considering moving the "spec" section (Init, Next and Liveliness) to the top (i.e. below constants and variables) to make it easier to understand the models for users. Kind of like an article where you read the highlights in the subheading.

What is generally recommended?

lemmy:
The definition of a formula has to come before its use.  In other words, the parser won't accept it.

nicholassm:
Makes sense.

Is there anything else you can see that I can do to improve this PR?
It's mostly related to (excellent) suggestions from @muenchnerkindl.

Kind regards,
Nicholas

nicholassm:
Hi guys,

Is there anything else I should do to increase the quality further? Otherwise, I think it's ready for merging. :-)
Many thanks in advance.

Kind regards
Nicholas

ahelwer (approved):

lemmy (approved):

PR #155, opened 2025-02-12
Finitizing monotonic systems Take2

lemmy:
Related to https://github.com/tlaplus/Examples/pull/97

Supersedes https://github.com/tlaplus/Examples/pull/153

PR #159, opened 2025-02-27
fixed proof of LoopInvariance/BinarySearch

muenchnerkindl:
Should now work with tlapm 1.6.0.


ahelwer:
This failed because PR https://github.com/tlaplus/Examples/pull/155 was not merged so we were using the non-rolling-release TLAPM version; could you resubmit it and also remove the failing proof from the skip-list in the CI file in .github/workflows/CI.yml?

PR #160, opened 2025-02-27
Update CI.yml

muenchnerkindl:
Remove FindHighest and BinarySearch from list of skipped proofs.


ahelwer:
This requires DCO signoff; you'll need to amend the commit to include the -S flag

muenchnerkindl:
Well, I tried, but apparently amended the wrong commit. Let's just forget about this one: the same changes are contained in some of the other PRs, so when everything is merged, everything should just work.

PR #170, opened 2025-05-21
Removed microwave submodule example from manifest.json

ahelwer:


lemmy:
Why remove?

ahelwer:
The other submodules were removed from CI checking for similar reasons. If we want the submodules to be included in the CI then they should be moved inside the repository. Possibly this would be less of an issue if we go to a devolved per-directory manifest.json file as suggested in #171.

lemmy:
> The other submodules were removed from CI checking for similar reasons. If we want the submodules to be included in the CI then they should be moved inside the repository. Possibly this would be less of an issue if we go to a devolved per-directory manifest.json file as suggested in #171, but then the people would be required to add a manifest.json file to their actual repository that is being submoduled.

There will always be spec that are referenced but not included directly in this repository for various reasons. Fortunately, Git submodules provide a relatively stable form of reference, as they capture not only the location of the external repo but also a specific commit in time. This minimizes the risk of CI brittleness—especially when the referenced spec is also hosted on GitHub.

In this case, the Microwave example was checked before this PR to ensure it works as expected for the *primary* audience of this repository. Therefore, please revert this PR.

ahelwer:
Perhaps we can discuss handling submodule specs as part of a more comprehensive change in #171. Your suggestions in https://github.com/tlaplus/Examples/pull/170#issuecomment-2898233027 of moving this to spec preamble/postamble would also not work with submodules as described, unless the origin repository is modified, in which case might as well bring the specs into this repo. It makes sense to unify these discussions instead of keeping them separate.

PR #175, opened 2025-06-30
Fix proofs

muenchnerkindl:
I went over all examples that contain proofs and made sure that they run with the current version of TLAPS, on both Linux and MacOS. Unfortunately, the CI terminates with error because of the use of an obsolete version of artefacts but after staring at CI.yml and some of the scripts, I still don't see where that happens.


lemmy:
Can you rebase on top of https://github.com/tlaplus/Examples/commit/1408f6c79fa4679ac73842d6124b5536883d36e9?

ahelwer:
(Have to search for ERROR:root in log to see error since there is quite a bit of non error output)

>INFO:root:specifications\ewd840\EWD840.cfg in 1.4s vs. 1s expected
2025-07-01T11:31:25.9099469Z ERROR:root:Recorded state count info differed from actual state counts:
2025-07-01T11:31:25.9100193Z ERROR:root:(distinct/total/depth); expected: (302, 2001, 9), actual: (302, 1, 9)

ahelwer:
Sorry! That CI run was canceled for some reason. It has been re-run and succeeds.

PR #187, opened 2025-08-21
Manifest: added proof runtimes

### LINKED ISSUE #123: Add theorem checking time to manifest details
Currently we manually exclude specific modules from proof checking in the CI because they run too long. Given that proof checking time can have as high of variability as model checking, it makes sense to include data about proof check time in the manifest then use that to determine whether to check the proof in the CI.

Failing proofs can continue to be manually excluded because ideally this repo would not have or accept any specs with failing proofs.

ahelwer:
Recorded runtime details for all existing proofs

Proofs running longer than 1 minute will not be run in CI

Closes #123


ahelwer:
Adding a `SMTT(30)` pragma to one troublesome proof in `VoteProof.tla` did help it finish under the timeout limit, good idea.

PR #192, opened 2025-12-17
CI: use bash instead of python for proof checking

ahelwer:
macOS runners also seem to be getting slower, so skip checking one of the CoffeeCan models.


ahelwer:
Closing this because the approach isn't quite there yet. I think I will need to implement a python script called something like is_proof.py that filters a pipeline of all .tla files found by `find`.

PR #198, opened 2026-01-19
Add "dag-based consensus"

nano-o:


nano-o:
Regarding my experience with the CI: it was good. One improvement could be to add a script that runs all the CI checks for a given specification directory (I ran them manually one by one).

PR #199, opened 2026-02-13
DieHardest: compare jug configurations via parallel and interleaved self-composition

lemmy:
Add DieHardest.tla, which composes two instances of DieHarder to compare which configuration has the shorter path to the Goal (state).

Four composition approaches are contrasted in clearly labeled sections that progressively motivate interleaved composition:

  1. Parallel — shows that BFS finds the shortest path for the slower configuration but not the faster one.
  2. Parallel + "per-behavior freeze" — shows that freezing after reaching the Goal does not help.
  3. Parallel + "global BFS-level freeze" — correct, but relies on TLC-specific operators.
  4. Interleaved — each instance steps independently, so BFS minimizes both step counts.

### DIFF
--- a/specifications/DieHard/DieHardest.tla
+++ b/specifications/DieHard/DieHardest.tla
@@ -1,0 +1,199 @@
+----------------------------- MODULE DieHardest ------------------------------
+(***************************************************************************)
+(* Given two jug configurations that can each solve the Die Hard problem,  *)
+(* which one reaches the Goal in fewer steps?                              *)
+(*                                                                         *)
+(* This is a question about pairs of behaviors — one from each             *)
+(* configuration — rather than about a single behavior.  Such properties   *)
+(* are called hyperproperties; ours is a 2-hyperproperty because it        *)
+(* relates two behaviors.  Ordinary model checkers like TLC check trace    *)
+(* properties (properties of individual behaviors), not hyperproperties.   *)
+(* The standard workaround is self-composition: run two copies of the      *)
+(* system side by side and reduce the hyperproperty to an ordinary         *)
+(* invariant of the product system.  This module does exactly that by      *)
+(* composing two instances of DieHarder.                                   *)
+(*                                                                         *)
+(* The choice of how the two copies advance matters.  Sections 1–3 show    *)
+(* that the natural choice — parallel (lock-step) composition — has a      *)
+(* subtle flaw: TLC's BFS finds the shortest trace for the slower          *)
+(* configuration but not necessarily for the faster one.  Two fixes are    *)
+(* explored; one works but departs from the logic of TLA+.  Section 4     *)
+(* gives a clean solution: interleaved composition, where each instance    *)
+(* steps independently.                                                    *)
+(*                                                                         *)
+(* Primary use case: show that adding a redundant jug (e.g. comparing      *)
+(* <<5, 3>> against <<5, 3, 3>> for Goal = 4) can shorten the solution.    *)
+(*                                                                         *)
+(* References:                                                             *)
+(*  - Lamport, "Verifying Hyperproperties With TLA", 2021.                 *)
+(*    (https://ieeexplore.ieee.org/document/9505222)                       *)
+(*  - Wayne, "Hypermodeling Hyperproperties", 2020                         *)
+(*    (https://hillelwayne.com/post/hyperproperties/).                     *)
+(***************************************************************************)
+EXTENDS Naturals, Functions, FiniteSetsExt, TLC
+
+(***************************************************************************)
+(* The Die Hard problem has a solution iff: (1) the Goal fits in the       *)
+(* largest jug, and (2) the Goal is divisible by the GCD of all jug        *)
+(* capacities.  Condition (2) follows from Bézout's identity: the          *)
+(* measurable quantities with jugs of capacities c1, …, cn are exactly     *)
+(* the multiples of GCD(c1, …, cn).  The LET definitions use distinct      *)
+(* names to avoid clashes with operators in DieHarder.                     *)
+(***************************************************************************)
+HasSolution(capacity, goal) ==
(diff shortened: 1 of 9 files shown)


lemmy:
@hwayne Could you please review this example? It references your “Hypermodeling Hyperproperties” article.

specifications/DieHard/DieHardest.tla
```diff
@@ -0,0 +1,192 @@
...
+(*  [j1=0, j2=10]   [j1=0, j2=3]                                           *)
+(*  [j1=9, j2=1]    [j1=1, j2=2]   c2 reaches Goal (2 steps) ←             *)
+(*  [j1=0, j2=1]    [j1=1, j2=2]   c2 frozen                               *)
+(*  [j1=1, j2=0]    [j1=1, j2=2]   c2 frozen                               *)
+(*  [j1=1, j2=10]   [j1=1, j2=2]   c2 frozen                               *)
+(*  [j1=9, j2=2]    [j1=1, j2=2]   c1 reaches Goal (6 steps) ←             *)
+(*                                                                         *)
+(* Both paths are individually shortest.  However, TLCSet and TLCGet       *)
+(* are TLC-specific operators outside the logic of TLA+, making this       *)
+(* approach incompatible with other TLA+ tools (e.g. Apalache).            *)
```
hwayne, line 158: Also, unless things changed recently, TLC global registers are local to each worker

lemmy: Indeed, this would also only work with a single worker.

hwayne:
These solutions depend on two assumptions: that the states are searched with BFS, and that the BFS is strict. The first is violated with the `-depth` and `-simulate` parameters. I've seen the second be violated when using multiple workers:

1. worker 1 picks the last state of length N on the queue
2. worker 2 picks the first state of length N+1
3. Both workers are evaluating a state that violates an invariant, but
4. worker 2 reaches the violation before worker 1 does.

It's pretty rare, but I've seen it happen!

A robust and universal TLA+ spec would probably have to store the diehard traces as data values and implement Dijkstra's algorithm to find the shortest path, with the property that `<>[](shortest_c1 <= shortest_c2)`.

lemmy:
> These solutions depend on two assumptions: that the states are searched with BFS, and that the BFS is strict. The first is violated with the `-depth` and `-simulate` parameters. I've seen the second be violated when using multiple workers:
> 
> 1. worker 1 picks the last state of length N on the queue
> 2. worker 2 picks the first state of length N+1
> 3. Both workers are evaluating a state that violates an invariant, but
> 4. worker 2 reaches the violation before worker 1 does.
> 
> It's pretty rare, but I've seen it happen!
>
>A robust and universal TLA+ spec would probably have to store the diehard traces as data values and implement Dijkstra's algorithm to find the shortest path, with the property that <>[](shortest_c1 <= shortest_c2).

Strict BFS has in fact been an implicit assumption in the current formulation. This directly contradicts the spec's own statement: "[_This approach stays entirely within the logic of TLA+ and is compatible with all TLA+ tools._](https://github.com/tlaplus/Examples/pull/199/changes#diff-c395f4e67c3ea608737395f9e906f1a75aca56fa0a376b7a8ec7d3624895d258R173-R174)" 

Given that contradiction, what do you think about removing that statement and instead making the assumption explicit, for example:

```tla
ASSUME TLCGet("config").mode = "bfs" /\ TLCGet("config").worker = 1
```

In my view, the example would still offer an interesting and valuable perspective on self-composition, even with this assumption made explicit.

hwayne:
> > These solutions depend on two assumptions: that the states are searched with BFS, and that the BFS is strict. The first is violated with the `-depth` and `-simulate` parameters. I've seen the second be violated when using multiple workers:
> > 
> > 1. worker 1 picks the last state of length N on the queue
> > 2. worker 2 picks the first state of length N+1
> > 3. Both workers are evaluating a state that violates an invariant, but
> > 4. worker 2 reaches the violation before worker 1 does.
> > 
> > It's pretty rare, but I've seen it happen!
> > A robust and universal TLA+ spec would probably have to store the diehard traces as data values and implement Dijkstra's algorithm to find the shortest path, with the property that <>[](shortest_c1 <= shortest_c2).
> 
> Strict BFS has in fact been an implicit assumption in the current formulation. This directly contradicts the spec's own statement: "[_This approach stays entirely within the logic of TLA+ and is compatible with all TLA+ tools._](https://github.com/tlaplus/Examples/pull/199/changes#diff-c395f4e67c3ea608737395f9e906f1a75aca56fa0a376b7a8ec7d3624895d258R173-R174)"
> 
> Given that contradiction, what do you think about removing that statement and instead making the assumption explicit, for example:
> 
> ```tla
> ASSUME TLCGet("config").mode = "bfs" /\ TLCGet("config").worker = 1
> ```
> 
> In my view, the example would still offer an interesting and valuable perspective on self-composition, even with this assumption made explicit.

I think this make sense, but then section 3's main downside (only TLC) then applies to all of the specs.

lemmy:
Added an Apalache module that, although slightly slower, produces the same counterexample for `NextInterleaved` as TLC. This confirms that the `NextInterleaved` approach is not specific to TLC.

That said, it is worth noting that TLA itself does not guarantee that a **shortest** counterexample will be found. Identifying shortest counterexamples is an implementation detail of particular verification tools, such as TLC and Apalache, rather than a property of the TLA language.

lemmy:
@ahelwer This PR enhances the GitHub automation to support overriding the `-workers` parameter.

ahelwer:
Nice! An alternative way to do it which is less directly bound to TLC would be to have an optional boolean parameter to the model in the manifest.json that is called something like `parallelizeStateExploration`. If this parameter is missing it defaults to true. Then TLC is run with either all the threads available on the machine or just in one thread. It is unlikely that someone would want to specifically run TLC in only two threads, for example. But this way you've done it is fine.

PR #209, opened 2026-04-22
Add Apalache wrappers and configurations for example specifications

lemmy:
AI has made substantial progress over the past few years, and that progress is starting to have very practical  implications for Apalache.

Around 2–3 years ago, I experimented with using AI to synthesize Apalache-style type annotations. At the time, even frontier models struggled to consistently produce syntactically correct annotations, which made the approach impractical.

Revisiting this recently with Claude Opus 4.7 shows a clear shift. When provided with the [type annotation guide](https://github.com/apalache-mc/apalache/blob/main/prompts/type-annotation-assistant.md) from the Apalache repository and given CLI access to Apalache (no Skills, MCPs, ...), the model was able to synthesize type annotations for a large portion of the existing TLA+ examples. The total wall-clock time for this process was on the order of a few hours.

That said, a meaningful portion of the remaining specs will require deeper changes. A key challenge is that Apalache's type system distinguishes between sequences, records, and functions, whereas in TLA+ these are all represented uniformly as functions. Bridging that conceptual gap will require structural refactoring rather than straightforward annotation, and [Apalache's Variant type](https://apalache-mc.org/docs/adr/002adr-types.html#13-type-system-12-including-precise-records-variants-and-rows) will likely be relevant.


ahelwer:
Good idea!

lemmy:
@ahelwer Please advise on f332e9d5b46a9f68efeb01983e0390b0e37ca731, b0fe6f7a87f71b98d9ba302e10458181bc36ccf7, and f07be26905e31f01569ecc26e3a5fc5e8d936cac

.github/scripts/tla_utils.py
```diff
@@ -137,6 +139,54 @@ def get_run_mode(mode):
...
+    re.IGNORECASE | re.MULTILINE,
+)
+
+def model_disables_deadlock_check(model_path):
+    """
+    Returns True if the given TLC config file contains a
+    `CHECK_DEADLOCK FALSE` directive (line-based, case-insensitive). Comment
+    handling is intentionally minimal: the directive must appear on its own
+    line. Used to translate the directive to Apalache's `--no-deadlock` CLI
+    flag, since Apalache silently ignores it in the config file.
```
konnov: It's worth opening an issue in Apalache on that

.github/workflows/CI.yml
```diff
@@ -45,23 +47,27 @@ jobs:
...
           python "$SCRIPT_DIR/check_manifest_features.py" \
             --examples_root .
       - name: Check README spec table
+        if: false # TEMP (Apalache PR): only run Apalache model checks.
         run: |
           python "$SCRIPT_DIR/check_markdown_table.py"  \
             --readme_path README.md
       - name: Convert specs to unicode
-        if: matrix.unicode
+        if: false && matrix.unicode # TEMP (Apalache PR): only run Apalache model checks.
```
konnov: You probably do not want this to land on main?

konnov (commented):
This looks nice! I really like the trick with introducing an additional instance to deal with type annotations. I noticed a few temporary fixes in the yml files. You probably want to fix those first

.github/scripts/tla_utils.py
lemmy: Is it feasible for this PR to wait on this feature to appear in Apalache?

.github/workflows/CI.yml
lemmy: No, this commit is specifically for this PR and will be dropped when this PR gets merged.

.github/scripts/tla_utils.py
lemmy: https://github.com/apalache-mc/apalache/pull/3312

lemmy:
> @ahelwer Please advise on [f332e9d](https://github.com/tlaplus/Examples/commit/f332e9d5b46a9f68efeb01983e0390b0e37ca731), [b0fe6f7](https://github.com/tlaplus/Examples/commit/b0fe6f7a87f71b98d9ba302e10458181bc36ccf7), and [f07be26](https://github.com/tlaplus/Examples/commit/f07be26905e31f01569ecc26e3a5fc5e8d936cac)

Your review comments have been incorporated.

.github/scripts/tla_utils.py
konnov: Thanks for submitting the patch. You can use the latest release now

lemmy:
The last commit introduces type annotations to a few additional steps.

Claude Opus 4.7 claims that it cannot add type annotations to the remaining specifications without altering them.

.github/scripts/tla_utils.py
```diff
@@ -163,16 +164,31 @@ def check_model(
...
+    # On Windows the launcher is `apalache-mc.bat`, and `subprocess.run`
+    # cannot exec batch files via CreateProcess directly, so route them
+    # through `cmd.exe /c`. On Linux/macOS it is the unsuffixed shell script
+    # `apalache-mc` and we invoke it directly.
+    if sys.platform == 'win32':
+        apalache_path = normpath(join(apalache_path, 'bin', 'apalache-mc.bat'))
+        apalache_launcher = ['cmd.exe', '/c', apalache_path]
+    else:
+        apalache_path = normpath(join(apalache_path, 'bin', 'apalache-mc'))
+        apalache_launcher = [apalache_path]
```
ahelwer: Python does have idiosyncratic scoping behavior where the apalache_launcher variable will be defined after this if/else block, but I do think it would be preferable to just assign the variable apalache_launcher at the same level as other variables in this block using a ternary operator.
```suggestion
    apalache_launcher = (
        ['cmd.exe', '/c', normpath(join(apalache_path, 'bin', 'apalache-mc.bat'))]
        if sys.platform == 'win32' else
        [normpath(join(apalache_path, 'bin', 'apalache-mc'))])
```

ahelwer (approved):
Looks good! Apalache will surely benefit from this increased testing, and users from the example of how Apalache type annotations can be added in a separate file.

PR #210, opened 2026-04-22
CI: invoke Apalache via apalache-mc.bat (and cmd.exe) on Windows

lemmy:
`tla_utils.check_model` hard-coded `bin/apalache-mc`, but on Windows the launcher is `bin/apalache-mc.bat` and `subprocess.run` cannot spawn `.bat` files via CreateProcess directly. Resolve the launcher per-platform and route batch files through `cmd.exe /c`. This was latent because the only symbolic-mode model in CI (EinsteinRiddle) was always being skipped: `[ ${{ matrix.unicode }} ]` in CI.yml is truthy for both "true" and "false", so the SKIP entry was added unconditionally rather than only on the Unicode variant.

### DIFF
--- a/.github/scripts/tla_utils.py
+++ b/.github/scripts/tla_utils.py
@@ -2,9 +2,11 @@ import locale
 from datetime import datetime
 from datetime import timedelta
 import json
+import os
 from os.path import join, normpath, pathsep, dirname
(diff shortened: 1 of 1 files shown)


lemmy:
Outdated

PR #211, opened 2026-04-23
Experiment with Claude Opus 4.7 and TLAPS.

lemmy:
Add specifications/ewd687a/EWD687a_proof.tla. All 408 obligations discharged by tlapm in ~38s (macbook pro M1).

Fully proved:

  THEOREM TypeCorrect            == Spec => []TypeOK
  THEOREM Thm_CountersConsistent == Spec => CountersConsistent

  via the combined inductive invariant Inv1 == TypeOK /\ Counters,
  where Counters relates the four per-edge counters:
    sentUnacked[e] = rcvdUnacked[e] + acks[e] + msgs[e].
  TypeOK alone is not inductive (RcvAck/SendAck decrement counters).

  LEMMA Inv2Inductive == Spec => []Inv2

  Inv2 is the structural overlay-tree strengthening of DT1Inv:
  non-neutral non-leader processes are in the upEdge tree, and
  upEdge[p] is a well-formed incoming edge of p with
  rcvdUnacked[upEdge[p]] >= 1. Inductiveness proved per action.

Left OMITTED (future work):

  LEMMA   DT1FromInv2      == Inv2 => DT1Inv          (chain/acyclicity)
  THEOREM Thm_DT1Inv       == Spec => []DT1Inv
  THEOREM Thm_TreeWithRoot == Spec => TreeWithRoot    (IsTreeWithRoot)
  THEOREM Thm_DT2          == Spec => DT2             (liveness)

Wire the module into CI: add a "proof" entry in manifest.json (picked up by the data-driven Check proofs step in .github/workflows/CI.yml) and flip the TLAPS Proof column in README.md.


lemmy:
Claude gave up on proving `Thm_TreeWithRoot`: 
> Prove Thm_TreeWithRoot

----

**Cursor**

Given the substantial time spent and 41 obligations still failing in my proof attempt, let me take a more pragmatic path. I'll revert `Thm_TreeWithRoot` back to `OMITTED` and explain the situation honestly.

Let me try one more focused approach to fix the recursive function definition issue.

Given the persistent difficulty getting `NatInductiveDef` to work with a locally-defined recursive function and the cascade of 41 failures, let me revert `Thm_TreeWithRoot` back to `OMITTED` and explain the situation.

I've made substantial progress restructuring the proof but have hit a fundamental TLAPS limitation: `NatInductiveDef` cannot be applied to a recursive function `iter` whose definition captures the outer `ASSUME` variable `q0` — even with the `HIDE DEF` + `Zenon` pattern that works in `LockHS.tla` (where the analogous function captures only constants like `{1}`).

Rather than burn more cycles, let me revert `Thm_TreeWithRoot` to its original `OMITTED` state but preserve the proof attempt as commented-out reference for future work, so we don't lose what's been built up.

The file has been reverted to the original state. Here is an honest summary of where things stand.

## Outcome

I was not able to prove `Thm_TreeWithRoot` in this session. After substantial work, I reverted `specifications/ewd687a/EWD687a_proof.tla` back to its committed state (with `Thm_TreeWithRoot` as `OMITTED`).

## What I built

The proof structure I developed proved most of the supporting machinery — what the theorem morally needs:

- `LEMMA ParentInTree` — strengthens `Inv2` to give parent properties for in-tree non-leaders.
- `LEMMA TWR_NodesInProcs`, `TWR_EInEdges` — graph-shape lemmas about the overlay tree's nodes/edges.
- `LEMMA TreeWithRootFromInvs` — the main reduction `Inv2 /\ NoCycle => TreeWithRootInv`, including subproofs for `IsDirectedGraph`, edge uniqueness, the `Transpose` calculation, and `AreConnectedIn` from the `Graphs` community module.
- The final `Thm_TreeWithRoot` would compose `Inv2Inductive`, `NoCycleInductive`, `TreeWithRootFromInvs`, and an `INSTANCE Graphs` unfolding step under `[]`.

## Where it broke

The proof depends on a small `LEMMA HasPath` that constructs a simple path from any in-tree non-leader `q0` to the leader by iterating `ParentStep`. The natural way to define that iteration is a recursive function on `Nat`:

```tla
iter[k \in Nat] == IF k = 0 THEN q0 ELSE ParentStep(iter[k-1])
```

and then justify it via `NaturalsInduction!NatInductiveDef` (the standard pattern used in `LockHS.tla`). In our setting `iter` captures the outer `ASSUME` constant `q0`, and TLAPS's second-order matching against `NatInductiveDefHypothesis` / `NatInductiveDefConclusion` / `NatInductiveDefType` consistently fails — even after trying:

- `HIDE DEF iter` + `BY Zenon DEF iter, NatInductiveDefHypothesis` (exact `LockHS` pattern),
- module-level `PathIter(q0)` helper with `LEMMA PathIterRec`,
- inline `LAMBDA v,k : ParentStep(v)` vs. named `Op(v,k)` vs. `Def(v,k)` operator,
- `Zenon`, `Isa`, and explicit re-stating of `NatInductiveDef` via local ASSUME/PROVE.

The same proof shape works in `LockHS.tla` because the recursive function there only captures the constant `{1}`. With a captured `q0` the obligation reduces to a second-order instantiation that none of the SMT/Zenon/Isabelle backends close in this configuration. After reaching this point the rest of `HasPath` (cardinality argument via `FS_PigeonHole` for the cycle-free iteration, prefix extraction) was working in the focused range tests, so the cascade of 41/1405 failures was driven by these few core obligations.

## Recommendation for finishing this off

Two reasonable next steps if you want to revisit:

1. **Use `RecursiveFcnOfNat`** (also in `NaturalsInduction.tla`) instead of `NatInductiveDef` — it expresses the existence of the unique recursive function more directly and may avoid the second-order matching path that's tripping the backends here.
2. **Lift the iteration out of `HasPath`** by adding it to `EWD687a.tla` itself (so `q0` becomes a parameter of a module-level operator without crossing an `ASSUME` boundary), and prove the `iter[0] / iter[k]` equation as a standalone module-level lemma.

I'm sorry I couldn't close it out within this session; the supporting lemmas above are sound and I can re-introduce them as a starting point whenever you'd like to take another pass.

specifications/byihive/VoucherLifeCycle_proof.tla
```diff
@@ -0,0 +1,19 @@
...
+(* TLAPS proof of                                                          *)
+(*    THEOREM VSpec => [](VTypeOK /\ VConsistent)                          *)
+(* stated in VoucherLifeCycle.tla.  TypeOK and VConsistent together form   *)
+(* an inductive invariant.                                                 *)
+(***************************************************************************)
+EXTENDS VoucherLifeCycle, TLAPS
+
+Inv == VTypeOK /\ VConsistent
+
+THEOREM Spec_TypeOK_Consistent == VSpec => [](VTypeOK /\ VConsistent)
```
muenchnerkindl: Since you define Inv above, consider using it on the right-hand side of the implication. You may also want to add USE DEF Inv, VTypeOK, VConsistent to the proof.

specifications/ewd687a/EWD687a_proof.tla
```diff
@@ -0,0 +1,859 @@
...
+(***************************************************************************)
+EXTENDS EWD687a, NaturalsInduction, FiniteSetTheorems, TLAPS
+
+-----------------------------------------------------------------------------
+(***************************************************************************)
+(* Re-state the (unnamed) ASSUME from EWD687a as a named assumption so we  *)
+(* can refer to it by name in proofs.  TLAPS does not currently expose     *)
+(* unnamed ASSUMEs to backends.                                            *)
+(***************************************************************************)
+ASSUME EdgeFacts ==
```
muenchnerkindl: Would it make sense to add the name to the ASSUME in EWD687a?

specifications/ewd687a/EWD687a_proof.tla
```diff
@@ -0,0 +1,859 @@
...
+(* non-leader p consists of non-neutral processes, so the leader cannot    *)
+(* be neutral whenever any other process is non-neutral.  Formalising the  *)
+(* finite-chain argument needs Procs to be finite and a small amount of    *)
+(* well-founded reasoning, factored out as a separate lemma.               *)
+(***************************************************************************)
+ASSUME ProcsFinite == IsFiniteSet(Procs)
+
+InTree(p) == upEdge[p] # NotAnEdge
+
+Inv2 == /\ TypeOK
```
muenchnerkindl: Instead of including TypeOK and Counters (which were already proved to be inductive) in the definition of Inv2, consider including Inv1 in the hypothesis of the preservation step and justify this by appealing to Inv1Inductive in the final PTL step of the invariant proof.

specifications/ewd687a/EWD687a_proof.tla
```diff
@@ -0,0 +1,859 @@
...
+(*                                                                         *)
+(* This requires reasoning about the IsTreeWithRoot predicate from the     *)
+(* community-modules Graphs module, in particular about the existence of   *)
+(* simple paths from every node to the root and the uniqueness of edges.   *)
+(* The structural invariant Inv2 above already captures the tree shape;    *)
+(* what is missing is the unfolding of IsTreeWithRoot/SimplePath/          *)
+(* Cardinality from the community-modules definitions, left as future      *)
+(* work.                                                                   *)
+(***************************************************************************)
+THEOREM Thm_TreeWithRoot == Spec => TreeWithRoot
```
muenchnerkindl, line 995: I may take a stab at the two following theorems when I have some time. No reason to wait for this to happen before pushing these changes.

specifications/ewd687a/EWD687a_proof.tla
```diff
@@ -0,0 +1,859 @@
...
+(* S' arose in the new state with p \in S', then by Counters and Inv2    *)
+(* conjunct 4 no other in-tree process points to p (since p was neutral, *)
+(* every OutEdge of p has sentUnacked = 0), so removing p from S' yields  *)
+(* a smaller closed set in the previous state - contradicting the         *)
+(* induction hypothesis.                                                   *)
+(***************************************************************************)
+NoCycle == \A S \in SUBSET (Procs \ {Leader}) :
+              ~ (/\ S # {}
+                 /\ \A q \in S : InTree(q) /\ upEdge[q][1] \in S)
+
```
muenchnerkindl, line 145: Some of the steps in the following proofs could certainly be simplified, for example by combining successive SUFFICES steps or chains of proof steps
``
<x>y. P
  BY ...
<x>y+1. Q
  BY <x>y
<x>y+2. R
  BY <x>y+1
``
where only R is used subsequently. But I don't think it's worthwhile doing that: as long as the proof is accepted, it's fine.

specifications/KeyValueStore/KeyValueStore_proof.tla
```diff
@@ -0,0 +1,43 @@
...
+(***************************************************************************)
+(* TLAPS proof of                                                          *)
+(*   THEOREM Spec => [](TypeInvariant /\ TxLifecycle)                      *)
+(* stated in KeyValueStore.tla.                                            *)
+(***************************************************************************)
+EXTENDS KeyValueStore, TLAPS
+
+Inv == TypeInvariant /\ TxLifecycle
+
+THEOREM TypeAndLifecycle == Spec => [](TypeInvariant /\ TxLifecycle)
```
muenchnerkindl: Again, consider writing "Spec => []Inv" in order to make use of the above definition, and perhaps include a USE step.

specifications/SpecifyingSystems/Composing/Channel_proof.tla
```diff
@@ -0,0 +1,14 @@
+--------------------------- MODULE Channel_proof ---------------------------
```
muenchnerkindl, line 1: The proofs in this directory appear to duplicate proofs in other subdirectories of SpecifyingSystems. But it probably make sense to leave this as it is so that the directories are aligned with the chapters of the book.

specifications/SpecifyingSystems/Liveness/HourClock_proof.tla
```diff
@@ -0,0 +1,14 @@
+-------------------------- MODULE HourClock_proof --------------------------
```
muenchnerkindl, line 1: I find it a little unfortunate that the following proofs are only about invariants whereas the chapter is called "Liveness". (Moreover, these proofs have already appeared elsewhere.) Aren't there any liveness properties asserted in the corresponding modules?

specifications/SpecifyingSystems/RealTime/HourClock_proof.tla
```diff
@@ -0,0 +1,14 @@
+-------------------------- MODULE HourClock_proof --------------------------
```
muenchnerkindl, line 1: Again, we have seen these proofs before, and there is no relationship to real time, contrary to what the chapter title suggests.

specifications/SpecifyingSystems/TLC/AlternatingBit_proof.tla
```diff
@@ -0,0 +1,67 @@
...
+  ASSUME NEW T, NEW s \in Seq(T), s # << >>
+  PROVE  Tail(s) \in Seq(T)
+  OBVIOUS
+
+LEMMA HeadType ==
+  ASSUME NEW T, NEW s \in Seq(T), s # << >>
+  PROVE  Head(s) \in T
+  OBVIOUS
+
+LEMMA LosePreservesType ==
```
muenchnerkindl, line 24: I would find it more natural to replace the existential quantifier in the hypothesis of this lemma by "NEW i \in 1 .. Len(q), " and drop step <1>1 in the proof. Also, since both main steps in the proof are proved by "OBVIOUS", the entire proof should probably be obvious? (Also, since after this change all "type" lemmas are OBVIOUS, can one drop their use in the invariant proof below and remove the type lemmas entirely? But sometimes it may be useful to state these lemmas anyway, in particular for the benefit of Zenon.)

specifications/transaction_commit/PaxosCommit_proof.tla
```diff
@@ -0,0 +1,31 @@
...
+(*   THEOREM PCSpec => PCTypeOK                                            *)
+(* stated in PaxosCommit.tla.                                              *)
+(*                                                                         *)
+(* Note that as written, the theorem only requires PCTypeOK at the initial *)
+(* state, since the right-hand side is a state predicate.  The strictly    *)
+(* stronger invariance result PCSpec => []PCTypeOK is more interesting     *)
+(* but requires an inductive strengthening for Phase2a (the chosen value   *)
+(* for the new phase 2a message must be in {"prepared","aborted"}, which   *)
+(* relies on the auxiliary invariant that phase1b messages with bal # -1   *)
+(* have val # "none").  That strengthening is left as future work.         *)
```
muenchnerkindl: Don't we have everything that is required here in the Paxos proof? (Perhaps look at tlapm/examples/paxos if the Paxos modules in the Examples collection do not contain all relevant proofs.)

specifications/transaction_commit/PaxosCommit_proof.tla
```diff
@@ -0,0 +1,31 @@
...
+(***************************************************************************)
+EXTENDS PaxosCommit, TLAPS
+
+(***************************************************************************)
+(* Pull the relevant facts out of the unnamed ASSUME in PaxosCommit.tla.   *)
+(***************************************************************************)
```
muenchnerkindl, line 258: Instead name the assumption in the base module?

specifications/transaction_commit/TCommit_proof.tla
```diff
@@ -0,0 +1,31 @@
...
+(***************************************************************************)
+(* TLAPS proof of                                                          *)
+(*   THEOREM TCSpec => [](TCTypeOK /\ TCConsistent)                        *)
+(* stated in TCommit.tla.                                                  *)
+(***************************************************************************)
+EXTENDS TCommit, TLAPS
+
+Inv == TCTypeOK /\ TCConsistent
+
+THEOREM TCorrect == TCSpec => [](TCTypeOK /\ TCConsistent)
```
muenchnerkindl: replace the conjunction on the right-hand side by Inv?

specifications/transaction_commit/TwoPhase_proof.tla
```diff
@@ -0,0 +1,40 @@
+--------------------------- MODULE TwoPhase_proof --------------------------
+(***************************************************************************)
+(* TLAPS proof of                                                          *)
+(*   THEOREM TPSpec => []TPTypeOK                                          *)
+(* stated in TwoPhase.tla.  TPTypeOK is directly inductive.                *)
+(***************************************************************************)
+EXTENDS TwoPhase, TLAPS
+
```
muenchnerkindl, line 23: How hard would it be to prove correctness of the protocol, i.e. no conflicting decisions?

muenchnerkindl (approved):
This looks good to me! A few detailed comments in the different modules, but nothing serious.

specifications/ewd687a/EWD687a_proof.tla
lemmy: Absolutely! This is a result of my prompting to prohibit modifications to the original spec to prevent the AI from cheating, such as altering the `Spec` to prove a theorem.

specifications/ewd687a/EWD687a_proof.tla
lemmy: Would you agree with the assumption that this will likely largely involve proving generic lemmas about graphs, and that we might want to add a graphs_theorems module to the CommunityModules/TLAPS?


Liveness proofs were mostly out-of-scope for this experiment, given the small corpus of available examples from which frontier models could have learned the technique.

specifications/transaction_commit/TCommit_proof.tla
lemmy: Going forward, I will add a corresponding instruction to my prompt since the AI frequently repeats this pattern.

specifications/ewd687a/EWD687a_proof.tla
muenchnerkindl: Yes, absolutely!

specifications/transaction_commit/TwoPhase_proof.tla
lemmy: https://github.com/tlaplus/Examples/pull/211/changes/2ade52697f0bc7ea019e6082dbbba3fd6fc654c6

specifications/ewd687a/EWD687a_proof.tla
lemmy: https://github.com/tlaplus/Examples/pull/211/changes/893baae02ff5f31d5bae8cf14f6c6163c559c1bf

PROOF_DIFFICULTY.md
```diff
@@ -0,0 +1,302 @@
...
+ascending effort.
+
+The survey covers all `*.tla` files under `specifications/` (excluding
+`.tlacache/` directories), 230 `THEOREM`s in total. Of those, 107 already
+carry a `BY` or structured proof. The remaining 123 — the focus of the
+ranking below — are either stated without proof or marked `OMITTED`.
+
+### Recent progress
+
+A first round added 18 Band-T/E proofs in companion `_proof.tla` files;
```
ahelwer: What is a Band-T/E proof?

lemmy: Trivial & Easy - The AI came up with it when tasked to categorize the proofs based on difficulty.

specifications/transaction_commit/PaxosCommit_proof.tla
lemmy: https://github.com/tlaplus/Examples/pull/211/changes/61963bc3d2ac7987f4ad29083e3d46754475da2f

specifications/SpecifyingSystems/TLC/AlternatingBit_proof.tla
lemmy: https://github.com/tlaplus/Examples/pull/211/changes/eaf34594ea0f69338f87bd67cf9885a4457a4240

PROOF_DIFFICULTY.md
ahelwer: It is helpful to define terms on first use.

lemmy: PROOF_DIFFICULTY.md is ephemeral and will be gone when the PR is ready for review.  Overall, this PR is to assess frontier model's capabilities (and perhaps gaps in TLAPS or the various _theorem.tla modules) when it comes to synthesizing TLAPS proofs.

lemmy:
@muenchnerkindl I had Claude tyr really ard but it ultimately failed to prove refinement: https://github.com/tlaplus/Examples/pull/211/changes/6f1ec2f0a194f2f1ea4128eda40dc840bbf9b534

lemmy:
I’m going to stop adding new commits for now until we figure out what to do with these results.

lemmy:
@muenchnerkindl You approved and earlier version of this PR.  I'll carry this over the finish line (squash commit, delete PROOF_DIFFICULTY, ...) if you also approve the most recent version.

muenchnerkindl:
Certainly, please go ahead! Thanks.

specifications/ewd687a/EWD687a.tla
muenchnerkindl: I suppose U should be E?

specifications/ewd998/EWD998_proof.tla
muenchnerkindl, line 56: Note that the following lemmas can (essentially) be found in https://github.com/tlaplus/CommunityModules/blob/master/modules/FunctionTheorems.tla, except that the premises corresponding to the three operators are inlined. (I agree that it is useful to have names for these properties.)

specifications/SpecifyingSystems/Liveness/HourClock_proof.tla
lemmy: I specifically instructed Claude to skip liveness proofs.

specifications/SpecifyingSystems/RealTime/HourClock_proof.tla
lemmy: Out of scope of this PR.

tangruize:
Hi, while investigating the bare QED behavior discussed in https://github.com/tlaplus/tlapm/issues/271 (a `<n> QED` without `BY`/`OBVIOUS` silently generates 0 obligations), we noticed two instances in `EWD998PCal_proof.tla` that look like they were accidentally left without `BY`:

https://github.com/tlaplus/Examples/blob/809c7a2d34a1a88a9aac6fabc6f92285aa6b0460/specifications/ewd998/EWD998PCal_proof.tla#L626
https://github.com/tlaplus/Examples/blob/809c7a2d34a1a88a9aac6fabc6f92285aa6b0460/specifications/ewd998/EWD998PCal_proof.tla#L714

`tlapm --summary` reports `missing_proofs_count = 2` for the theorem at line 533.

(BTW, we compiled a checklist of common proof anti-patterns like this one — particularly relevant when using AI to generate proofs: [PROOF_ANTIPATTERNS.md](https://github.com/tangruize/tlaps-bench/blob/master/PROOF_ANTIPATTERNS.md))

lemmy:
@tangruize Addressed in https://github.com/tlaplus/Examples/commit/3a6b92c4593a9fd43ca9bd885854c40c3d3c1ac5.  Thank you!

lemmy:
> (BTW, we compiled a checklist of common proof anti-patterns like this one — particularly relevant when using AI to generate proofs: [PROOF_ANTIPATTERNS.md](https://github.com/tangruize/tlaps-bench/blob/master/PROOF_ANTIPATTERNS.md))

Thanks for sharing PROOF_ANTIPATTERNS.md and the associated Python scripts.

I ran the script across this repo with #211 and #212 merged, and it didn’t turn up any smoking guns. Moving the proof into a _proof.tla file seems like a good first line of defense. On the other hand, it may simply be that Codex has a stronger tendency to cheat than Claude Opus 4.7.

PR #215, opened 2026-07-18
safety proof of EWD687a

muenchnerkindl:
Updated EWD687a according to the changes in the Graphs community module, and contributed a proof of the safety property of the algorithm.

### DIFF
--- a/specifications/ewd687a/EWD687a.tla
+++ b/specifications/ewd687a/EWD687a.tla
@@ -377,13 +377,13 @@ RcvMsg(p) == \E e \in InEdges(p) :
                   /\ UNCHANGED <<acks, sentUnacked>>
  
 (***************************************************************************)
-(* A process p may finish its computation and become idle at any time.     *)
+(* An active process p may become idle at any time.                        *)
 (*                                                                         *)
-(* If a non-leader process p is neutral after an idle step, it implies     *)
-(* that p was not a node of the overlay tree when it became idle.  Thus,   *)
-(* there is no need to change upEdge[p] in the  Idle  subaction.           *)
+(* Since p is active before the step, it must be a node of the overlay     *)
+(* tree. If it is a leaf, it can be removed from the tree in a subsequent  *)
+(* SendAck transition.                                                     *)
 (***************************************************************************)
-Idle(p) == /\ active[p]  \* otherwise this step is subsumed by stuttering
+Idle(p) == /\ active[p]
            /\ active' = [active EXCEPT ![p] = FALSE]
            /\ UNCHANGED <<msgs, acks, sentUnacked, rcvdUnacked, upEdge>>
 
@@ -419,9 +419,7 @@ THEOREM Spec => []CountersConsistent
 
 TreeWithRoot ==
     LET E == {upEdge[p] : p \in DOMAIN upEdge} \ {NotAnEdge}
-        \* The original definition includes every node occurring in the tree
-        \* twice (except the leader), which unnecessarily complicates the proof.
-        N == (* {e[1] : e \in E} \cup *) {e[2] : e \in E} \cup {Leader}
+        N == {e[2] : e \in E} \cup {Leader}
         O == Transpose([edge |-> E, node |-> N])
     IN \* O is a tree rooted in the leader.
        /\ IsTreeWithRoot(O, Leader)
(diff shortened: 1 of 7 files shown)


specifications/ewd687a/EWD687a.tla
```diff
@@ -383,16 +383,17 @@ RcvMsg(p) == \E e \in InEdges(p) :
 (* that p was not a node of the overlay tree when it became idle.  Thus,   *)
 (* there is no need to change upEdge[p] in the  Idle  subaction.           *)
 (***************************************************************************)
-Idle(p) == /\ active' = [active EXCEPT ![p] = FALSE]
+Idle(p) == /\ active[p]  \* otherwise this step is subsumed by stuttering
```
lemmy: I suggest elaborating on why this is an issue.

specifications/ewd687a/EWD687a.tla
```diff
@@ -401,34 +402,33 @@ Spec == Init /\ [][Next]_vars /\ WF_vars(Next)
...
     LET E == {upEdge[p] : p \in DOMAIN upEdge} \ {NotAnEdge}
-        N == {e[1] : e \in E} \cup {e[2] : e \in E}
-        G == INSTANCE Graphs
-        O == G!Transpose([edge |-> E, node |-> N])
-    IN [](
-          \* O is a tree rooted in the leader.
-          /\ O.edge # {} => G!IsTreeWithRoot(O, Leader)
-          \* All nodes of the overlay tree are non-neutral.
-          /\ N:: \A n \in O.node: ~neutral(n))
+        \* The original definition includes every node occurring in the tree
```
lemmy: The comment about the original definition can likely be omitted.

specifications/ewd687a/MCEWD687a.cfg
```diff
@@ -1,45 +1,25 @@
-\* MV CONSTANT declarations
 CONSTANTS
+    \* model values for processes
     L = L
     P1 = P1
     P2 = P2
     P3 = P3
-\* MV CONSTANT definitions
+    MaxCounter = 3
+    Leader = L
```
lemmy: Only keep `Leader` or `L`?

lemmy (approved):
LGTM

lemmy:
Can the reduction of proof effort be soley attributed to the introduction of GraphTheorems?

specifications/ewd687a/EWD687a.tla
muenchnerkindl: It would require an extra case distinction in the proof of the tree invariant. But your remark prompted me to adapt the comment explaining the Idle action.

muenchnerkindl:
> Can the reduction of proof effort be soley attributed to the introduction of GraphTheorems?

In fact, the previous proof did not establish DT1Inv (i.e., correctness of the algorithm). It did show that the overlay graph was acyclic, but a dag is not quite enough, one really needs a tree. It had also suggested a strengthening of DT1Inv by adding an additional conjunct to acyclicity. Replacing the acyclicity condition with the tree predicate was enough for the proof to go through, based on the lemmas about graphs. Also, I took inspiration from the way acyclicity was expressed for proving a general lemma that implies that a tree is acyclic. Overall, the work done by Claude was very useful, even it didn't quite manage to prove overall correctness.

A technicality was that the tree invariant made use of an instance introduced in a `LET` construct, which I would not know how to explain to the PM. I'll take this into account when revisiting the handling of `INSTANCE` by the PM, but dropping `LET`-bound instances would perhaps also be a candidate for simplifying the language?

specifications/ewd687a/EWD687a.tla
muenchnerkindl: Incidentally, I noticed that you had added an action constraint to the model used for animation precisely to rule out such stuttering steps caused by the `Idle` action. I removed that because it has become unnecessary with the changed definition.

lemmy:
> ...dropping `LET`-bound instances would perhaps also be a candidate for simplifying the language?

That would cripple the current TLC REPL, which relies solely on `LET`-bound instances to load modules due to the absence of EXTENDS support.

```tla
> % tlcrepl
Welcome to the TLA+ REPL!
TLC2 Version 2026.04.27.203507
Enter a constant-level TLA+ expression.
(tla+) LET B == INSTANCE Bags IN B!SetToBag({})
<<>>
```

muenchnerkindl:
> > ...dropping `LET`-bound instances would perhaps also be a candidate for simplifying the language?
> 
> That would criplle the current TLC REPL, which relies solely on `LET`-bound instances to load modules due to the absence of EXTENDS support.

Understood.

PR #217, opened 2026-07-29
Factor the TypeOK inductive step out as LEMMA TypeOKStep

lemmy:
SafetyStep re-derived TypeOK /\ [Next]_vars => TypeOK' inline: the step existed only inside the proof of THEOREM TypeCorrect, and proof steps are not citable from another theorem, while Spec => []TypeOK is temporal and SafetyStep assumes no temporal hypothesis.

Hoist the step to a top-level lemma; TypeCorrect's <1>2 and SafetyStep's <1>typok now both cite it. The proof body is unchanged, step levels included. tlapm --nofp proves all 199 obligations.

### DIFF
--- a/specifications/ReadersWriters/ReadersWriters_proof.tla
+++ b/specifications/ReadersWriters/ReadersWriters_proof.tla
@@ -42,10 +42,7 @@ LEMMA TailIsSeq ==
 (***************************************************************************)
 (* Type correctness.                                                       *)
 (***************************************************************************)
-THEOREM TypeCorrect == Spec => []TypeOK
-<1>1. Init => TypeOK
-  BY DEF Init, TypeOK
-<1>2. TypeOK /\ [Next]_vars => TypeOK'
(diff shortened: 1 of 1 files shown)


specifications/ReadersWriters/ReadersWriters_proof.tla
muenchnerkindl: Sure, it is nonsense to copy the proof of type preservation in this proof. Factoring out this step as an extra lemma is one solution. The one that I tend to use, given that ``Spec => []TypeOK`` has already been established, is to set up the subsequent invariant proof as follows:
```
THEOREM Spec => []Safety
<1>1. Init => Safety
<1>2. TypeOK /\ TypeOK' /\ Safety => Safety'
<1>. QED
  BY <1>1, <1>2, TypeCorrect, PTL DEF Spec
```
The extra hypotheses `TypeOK` and `TypeOK'` are justified by the type-correctness theorem. This saves you from defining the extra predicate `Inv` as the conjunction of the two invariants.

PR #218, opened 2026-08-01
EWD998: close the nine lemmas that were stated without proof

vasilisnasopoulos:
Follow-up to the discussion in tlaplus/CommunityModules#128.

`EWD998_proof.tla` states nine lemmas with no proof at all. Steps carrying no proof generate no obligation, so `tlapm` reports *All 807 obligations proved* and exits 0 — while `--strict` (tlaplus/tlapm#278) exits 11. This closes all nine.

**The six `FoldFunctionOnSet*` lemmas** are removed and imported from `FunctionTheorems`, where they are proved. Their statements are the same up to generalisation; the community versions have weaker hypotheses.

**`SumIsInt`, `SumIsNat`, `SumEqual`** get structured proofs on top of those, using `FS_Subset` to carry finiteness from `Node` down to the index set — which is what the old `BY` lines were missing.

**`SumIterate` and `SumUnion`** do not go through that route. TLAPS cannot instantiate the second-order `op(_,_)` against the `LAMBDA x1, x2 : x1 + x2` inside `Sum`; the file already carried commented-out attempts marked `(* fails *)`. They are proved instead from `MapThenSumSetAddElement` and `MapThenSumSetDisjointUnion` in `FiniteSetsExtTheorems`, whose operator is *unary*, and which coincide with `Sum` once the definitions are unfolded. Thanks to @muenchnerkindl for the pointer.

**`Functions.tla`** gains `SumFunctionOnSet` and `SumFunction` (10 added lines, nothing removed). `FunctionTheorems` needs them and the vendored copy predates them.

**`manifest.json`**: `maxRuntimeMinutes` 2 → 4. Locally the module went from about a minute to 1m33s with `--stretch 5` on a cold cache.

## Result

| | before | after |
|---|---|---|
| lemmas stated without proof | 9 | **0** |
| `--strict` exit | 11 | **0** |
| obligations | 807 | **851, all proved** |

Verified with `tlapm 1.6.0-pre` (`4600b24`), invoked as the CI does:

```sh
tlapm --strict specifications/ewd998/EWD998_proof.tla -I <community> --stretch 5
```

`AsyncTerminationDetection_proof.tla` in the same directory is unaffected (exit 0, 30 obligations). `EWD998PCal_proof.tla` fails for me with `Unknown module "BagsExt"` both before and after this change, so that is my include path rather than anything here.


specifications/ewd998/EWD998_proof.tla
muenchnerkindl, line 45: If I am not mistaken, these operators aren't used anywhere in this module. If that is indeed so, I suggest removing what follows up until `LEMMA PlusACI`, except for lemma `NodeIsFinite`, which is indeed required.

specifications/ewd998/Functions.tla
```diff
@@ -6,6 +6,7 @@
 (*  \vspace{12pt}}^'                                                       *)
 (***************************************************************************)
 
+EXTENDS Integers
```
muenchnerkindl: Since this is a subset of module `Functions` of the Community Modules, I suggest removing this copy of the module: we anyway import some of the Community Modules, and there is no reason to have a local copy here. (This is probably historical: the module originated from within the TLAPS distribution and was later migrated to the Community Modules and subsequently extended.)

muenchnerkindl (commented):
Thanks a lot for this quick fix! I only have two suggestions on removing operators that now have become irrelevant and on removing the local copy of the Functions module, which seems irrelevant.

NB: The BagsExt module is also part of the Community Modules. Also note that EWD998PCal_proof.tla contains a theorem whose proof is OMITTED, but this is documented in a comment.

vasilisnasopoulos:
Both applied, thank you.

`Functions.tla` is gone — you are right that there is no reason for a local copy once the Community Modules are on the include path. `EWD998.tla` and `Utils.tla` are the only modules here that extend it, and both resolve fine against the community version.

On the operators: `IsAssociativeOn`, `IsCommutativeOn` and `IsIdentityOn` existed only to state `PlusACI`, and `PlusACI` itself is no longer referenced anywhere now that the fold lemmas come from `FunctionTheorems` — the new proofs state the algebraic side conditions inline. So all four are removed rather than just the three; `NodeIsFinite` stays.

850 obligations, all proved, `--strict` exit 0.

Also confirming your two notes, now that I have `BagsExt` on the path: `EWD998PCal_proof.tla` runs and proves all 1180 obligations, exiting 11 for the documented `OMITTED` theorem — so my earlier report of exit 3 there was purely my missing include path. `AsyncTerminationDetection_proof.tla` is unaffected at exit 0.

muenchnerkindl (approved):
Thanks, all looks good to me! Merging into master.

vasilisnasopoulos:
Sorry, that was my fault: I removed `Functions.tla` but left its entry in `manifest.json`, so `check_manifest_files.py` failed everywhere. Fixed in `cab9d84`.

I reproduced the check locally this time — both `.tla` and `.cfg` paths, in both directions, across all manifests — and it comes out clean. The `EWD998 Trace Validation` workflow had already passed on the previous run.

PR #220, opened 2026-08-04
Add Vortex DSE specifications

vasilisnasopoulos:
Adds `specifications/VortexDSE/` — a deterministic slot-based admission
protocol with a separate per-slot agreement layer. A message carries its own
slot stamp and each node decides admission locally against its own clock;
there is no leader, quorum or vote.

**Two admission modes, both specified because both are implemented.**
`Vortex_DSE_CSlot` admits a late message into its own slot, so nothing is
dropped. `Vortex_DSE_CSlot_TTL` rejects it permanently, which bounds memory
behind the frontier. The difference is `<=` versus `=` in one conjunct.

`Vortex_DSE_CSlot_AE` (Freeze / Reconcile / Commit) is specified over the TTL
mode, not the default. Its `NoReorderAcrossCslot` invariant encodes the strict
gate, so it is not a refinement of the late-tolerant rule; extending it would
mean restating what "no reordering across slots" means, and I have not done
that here. The spec README says this rather than leaving it to be discovered.

**Checked:**

| | |
|---|---|
| TLC models | 6, all `success`, 0.6–3.3s |
| `Vortex_DSE_CSlot_Proofs` | 194 obligations |
| `Vortex_DSE_CSlot_ExactlyOnce_Proof` | 131 obligations |
| `Vortex_DSE_CSlot_AE_Proofs` | 34 obligations |

All three proof modules exit 0 under `tlapm --strict` and contain no `OMITTED`
steps. Model runtimes and state counts in `manifest.json` were measured with
`check_small_models.py` locally; `check_manifest_schema`, `check_manifest_files`,
`check_manifest_features` and `check_markdown_table` all pass.

Scope is deliberately limited: `Vortex_DSE_CSlot_Skew` states the bounded-skew
assumption structurally but does not model the mechanism that maintains it, and
`Reconcile` is a single atomic step rather than the multi-round protocol
underneath.

Two notes from preparing this, neither blocking:

- `generate_manifest.py` rewrites every spec's `manifest.json`, not only the new
  one; it produced diffs in 14 unrelated manifests, which I reverted. Only
  `specifications/VortexDSE/` and the README table are touched here.
- `check_proofs.py` reads `module['proof']['runtime']`, but the schema requires
  `maxRuntimeMinutes` and sets `additionalProperties: false`. It raises
  `KeyError: 'runtime'` on any spec with a proof, including on a clean `master`.
  CI is unaffected because it invokes `tlapm` directly. Happy to open a separate
  issue or PR for it if useful.


specifications/VortexDSE/MC_Vortex_DSE_CSlot_AE.tla
```diff
@@ -0,0 +1,22 @@
...
+
+EXTENDS Vortex_DSE_CSlot_AE
+
+ConstInit ==
+    /\ Nodes   = {"n1", "n2"}
+    /\ MsgIDs  = {"a", "b"}
+    /\ MaxSlot = 1
+
+\* Conjunction of every safety invariant in the module.
+AllInv ==
```
lemmy: Isn’t it better if Apalache reports the name of the invariant that’s been violated?

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+
+EXTENDS Naturals, FiniteSets
+
+CONSTANTS
+    \* @type: Set(Str);
+    Nodes,           \* finite set of node identifiers
+    \* @type: Set(Str);
+    MsgIDs,          \* finite set of distinct message identifiers
+    \* @type: Int;
+    MaxSlot          \* slot horizon (state-space bound)
```
lemmy: MaxSlot is a model-checking concern, so it doesn’t belong in the high-level specification. Instead, TLC should be used to box the state space by using state and action constraints.

lemmy:
If any part of this commit was generated with AI assistance, please add a `Co-authored-by` footer naming the model to the commit message. AI use is permitted; we simply ask that it be disclosed for transparency.

lemmy:
> Two notes from preparing this, neither blocking:
> 
> * `generate_manifest.py` rewrites every spec's `manifest.json`, not only the new
>   one; it produced diffs in 14 unrelated manifests, which I reverted. Only
>   `specifications/VortexDSE/` and the README table are touched here.
> * `check_proofs.py` reads `module['proof']['runtime']`, but the schema requires
>   `maxRuntimeMinutes` and sets `additionalProperties: false`. It raises
>   `KeyError: 'runtime'` on any spec with a proof, including on a clean `master`.
>   CI is unaffected because it invokes `tlapm` directly. Happy to open a separate
>   issue or PR for it if useful.

@ahelwer ^^

specifications/VortexDSE/MC_Vortex_DSE_CSlot_AE_liveness.cfg
```diff
@@ -0,0 +1,13 @@
+SPECIFICATION LiveSpec
+
+CONSTANTS
+    Nodes   = {n1, n2}
+    MsgIDs  = {m1}
+    MaxSlot = 1
+
+PROPERTIES
```
lemmy, line 10: Careful when checking liveness properties under state- and action-constraints.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+\* processed AND the slot the message belongs to has been reached
+\* (m.cslot <= current_slot). The message keeps its own C_slot.
+\* Late delivery (m.cslot < current_slot) is ADMITTED, not dropped: it is
+\* placed into its own (earlier) slot. Nothing is lost.
+\* Future-dated (m.cslot > current_slot) waits: it cannot be admitted
+\* before the ticker reaches its slot (that slot has not happened yet).
+Process(n, m) ==
+    /\ n \in Nodes
+    /\ m \in network
+    /\ node_state[n] = "up"
```
lemmy: nit: consider extracting strings like "up", ... into constants.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+\* Attacker injects a message with arbitrary cslot value (past, present,
+\* or future). The C-slot gate must still hold.
+DuplicateInject(id, fake_cslot) ==
+    /\ id \in MsgIDs
+    /\ fake_cslot \in 0..MaxSlot
+    /\ network' = network \cup {[id |-> id, cslot |-> fake_cslot]}
+    /\ UNCHANGED <<current_slot, processed, persisted, node_state>>
+
+\* Slot ticker advances by 1.
+Tick ==
```
lemmy, line 121: Related: https://github.com/tlaplus/Examples/tree/master/specifications/FiniteMonotonic

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+\* not on n. Same (m.cslot, current_slot) => same decision at every node.
+DecisionLocalityOnly ==
+    \A n1, n2 \in Nodes : \A id \in MsgIDs :
+        (id \in processed[n1] /\ id \in processed[n2]) =>
+            (\E m \in network : m.id = id)
+
+-------------------------------------------------------------------------------
+(*                          STATE-SPACE CONSTRAINT                          *)
+
+StateConstraint ==
```
lemmy: This and the constant `MaxSlot` should be pushed into one of the MC_...tla files.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+(*    Process; SF guarantees that a message enabled infinitely often is     *)
+(*    eventually admitted. This is what recovers VALIDITY: every TX that    *)
+(*    reaches the network is eventually admitted by every up node.          *)
+(*                                                                          *)
+(*  - NO fairness on Submit / DuplicateInject. Submit is a user action;    *)
+(*    adversary injection is, by definition, not fair.                     *)
+(***************************************************************************)
+
+Fairness ==
+    /\ SF_vars(Tick)
```
lemmy: WF_ should be sufficient.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+    /\ \A n \in Nodes : WF_vars(Rejoin(n))
+    /\ \A n \in Nodes : SF_vars(\E m \in network : Process(n, m))
+
+LiveSpec == Init /\ [][Next]_vars /\ Fairness
+
+\* L1 TICK PROGRESS.
+\* Under SF(Tick), the slot counter eventually reaches the horizon.
+TickProgress == <>(current_slot = MaxSlot)
```
lemmy: This could be strengthened to <>[], but `MaxSlot` should be here anyway.

specifications/VortexDSE/Vortex_DSE_CSlot_AE.tla
```diff
@@ -0,0 +1,270 @@
+-------------------- MODULE Vortex_DSE_CSlot_AE --------------------
+(***************************************************************************)
+(* Vortex DSE — Agreement Extension Layer (L4)                              *)
+(*                                                                          *)
+(* Companion module to Vortex_DSE_CSlot.tla. The core module models the     *)
```
lemmy, line 5: Should this module refine the other (under some suitable refinement mapping)?

vasilisnasopoulos:
Thank you — this was a more careful read than I expected, and the `MaxSlot` point was right.

`MaxSlot` is gone from all four specifications. Each now has an `MC_` module that fixes the horizon **inside the actions** rather than through a `CONSTRAINT`, which also answers the liveness warning: no successor states are discarded, so the temporal properties are checked against a genuinely finite graph rather than a truncated one. There is no `CONSTRAINT` left anywhere in the directory. `MaxSkew` stays in the Skew module, since it is an assumption the protocol relies on rather than a checking artifact.

`FiniteMonotonic` was the useful pointer — the `MC_` modules follow its shape.

Also done: `WF_vars(Tick)` now that the ticker is unbounded; `TickProgress` strengthened to `<>[]` in the MC modules, where the horizon makes it permanent; node states and AE phases named rather than written as bare strings; and `AllInv` dropped so a checker reports which invariant failed.

On refinement — a fair question, and the answer today is no.

The obstacle is not the admission rule. It is that `Vortex_DSE_CSlot_AE` does something neither core module can: `NextCslot` clears `processed` for **every** node at once at a slot boundary. In the core specifications `processed` only grows, except when a single node crashes and goes down. So AE behaviours are not core behaviours under any mapping that hides `phase` and `committed_set`. The two also model disjoint failure dimensions — the core has crash and rejoin, AE has neither, deliberately.

They are complementary views rather than stacked layers: the core is accumulation across slots under crash, AE is one slot's agreement cycle with a reset at the boundary. The README now says as much instead of implying composition.

A module that does cross-cut both is the right thing to have, and I intend to write it — but as its own contribution, with a proved mapping, rather than asserted here.

Seven TLC models, all under five seconds. The three proof modules discharge 191, 128 and 32 obligations and exit 0 under `tlapm --strict`.

vasilisnasopoulos:
The macOS failure here is not from this change. `EWD998_proof.tla` hit its 4-minute wall in the shared proof step:

```
54195 Killed: 9  timeout --signal=KILL 4m ... EWD998_proof.tla
```

It proved fine in the other five jobs, the proof step runs sequentially so the modules added here do not compete for CPU, and they take about three seconds between them.

That budget is one I set in #218, and it is too tight — locally the proof takes 86 seconds, while the macOS runner needed more than 2.8x that. #221 raises it to 6 minutes.

vasilisnasopoulos:
Added the `Co-authored-by` footer to both commits here, and to #221. Claude Opus 5 throughout — writing the module and README text, running the models and proofs, and preparing the manifest.

lemmy:
Build problem appears to be caused by https://github.com/apalache-mc/apalache/pull/3432#issuecomment-5194573693

lemmy:
FWIW: https://github.com/tlaplus/vscode-tlaplus/blob/master/resources/knowledgebase/tla-review-guidelines.md might be useful for your agentic workflow

vasilisnasopoulos:
The red here is not from this branch. `master` itself is failing all six jobs right now — [run 31018433496](https://github.com/tlaplus/Examples/actions/runs/31018433496) — for the same reasons:

```
specifications/GermanProtocol/APGermanData.cfg      expected success, got 1
specifications/FlashProtocol/APFlashWithMutex.cfg   expected success, got 1
Error: Could not create the Java Virtual Machine.
```

Apalache is not starting on the runners. This branch additionally hits `smoke_test_large_models.py`, where TLC reports `SmokeEWD998.cfg: File not found` although the file is present in the tree — a working-directory problem rather than a missing file.

Nothing under `specifications/VortexDSE/` appears in either failure, and its own models and proofs pass in every job.

vasilisnasopoulos:
Thanks for the guidelines link — I ran the specs against it and four things were not being followed.

**Constants had no assumptions.** All four specifications declared `Nodes` and `MsgIDs` with nothing constraining them, and Skew's `MaxSkew` with nothing saying it is a natural number. Named assumptions are in now. Worth having here in particular, since these modules carry TLAPS proofs.

**Type invariants** are back to the `var \in S` form — `network \subseteq MsgRecord` rather than a quantifier over its elements. `MsgRecord` ranges over `Nat` and cannot be enumerated, so the MC module carries a bounded `MCTypeInvariant` for TLC instead of the specification bending to suit it.

**Strong fairness on Process** is now justified by measurement rather than assertion. I tried weak fairness, as the guidelines ask: the liveness model reports a temporal-property violation under it, because a crash intermittently disables `Process`. That is recorded next to the definition.

**One deviation, deliberate.** The guidelines prefer a state `CONSTRAINT` over enablement conditions, and I moved the horizon into `MCTick` instead. I tried the constraint first and it does not work here: TLC evaluates invariants on the state that crosses the boundary before the constraint discards it, so with `MaxSlot = 2` a `Tick` reaches `current_slot = 3` and any invariant mentioning the horizon fails there. The reason is in the module header rather than left unexplained. Happy to be told there is a better way around it.

Seven TLC models, all under five seconds; the three proof modules discharge 191, 128 and 32 obligations and exit 0 under `tlapm --strict`.

vasilisnasopoulos:
The refinement module is written and proved, on a branch stacked on this one. It keeps the core's variables, lets `processed` accumulate rather than clearing it at a slot boundary, and holds the per-slot input set in `committed`, so `THEOREM Refinement == Spec => C!Spec` goes through under the identity mapping — 25 obligations, exit 0 under `tlapm --strict`, and TLC checks it as a temporal property as well.

I will open it as its own pull request once this one lands, rather than growing this one further.

lemmy:
master is green again

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+    /\ n \in Nodes
+    /\ node_state[n] = Down
+    /\ processed'  = [processed  EXCEPT ![n] = persisted[n]]
+    /\ node_state' = [node_state EXCEPT ![n] = Up]
+    /\ UNCHANGED <<current_slot, network, persisted>>
+
+\* Adversarial duplicate / replay injection.
+\* Attacker injects a message with arbitrary cslot value (past, present,
+\* or future). The C-slot gate must still hold.
+DuplicateInject(id, fake_cslot) ==
```
muenchnerkindl: This action clearly subsumes Submit(id), why do you have both – in particular given that there is no fairness constraint on Submit?

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+    /\ processed    \in [Nodes -> SUBSET MsgIDs]
+    /\ persisted    \in [Nodes -> SUBSET MsgIDs]
+    /\ node_state   \in [Nodes -> {Up, Down}]
+
+-------------------------------------------------------------------------------
+(*                       CORE SAFETY INVARIANTS                             *)
+
+\* I1: EXACTLY-ONCE PER NODE.
+\* No node processes the same id twice (set semantics + guard).
+ExactlyOncePerNode ==
```
muenchnerkindl: This is a direct consequence of the type invariant. I think you even have the stronger invariant
``\A n \in Nodes : processed[n] \subseteq {m.id : m \in network}``
which states that only sent messages are processed.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+    \A n \in Nodes : Cardinality(processed[n]) <= Cardinality(MsgIDs)
+
+\* I2: NO FUTURE ADMISSION (the headline safety property).
+\* A node never admits a message whose slot has not yet been reached.
+\* The gate is m.cslot <= current_slot and current_slot is monotonic,
+\* so every admitted id has a network record whose cslot lies in
+\* [0, current_slot]: a real, present-or-past slot, never future-dated.
+\* Late messages (cslot < current_slot) ARE admitted here (into their
+\* own slot) — that is intended; only future-dated admission is barred.
+NoFutureAdmission ==
```
muenchnerkindl, line 159: NB: The invariant I suggested above is a consequence of this one.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+    \A n \in Nodes : \A id \in processed[n] :
+        \E m \in network : m.id = id /\ m.cslot <= current_slot
+
+\* I3: PERSISTED REFLECTS REALITY.
+\* mmap snapshot never invents ids that were not in the network.
+PersistedReflectsReality ==
+    \A n \in Nodes :
+        node_state[n] = Down =>
```
muenchnerkindl: Why is this asserted only when the node is down? It should hold at all times?

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+            persisted[n] \subseteq {m.id : m \in network}
+
+\* I4: NO PHANTOM PROCESS.
+\* Every processed id corresponds to a real network record.
+NoPhantomProcess ==
```
muenchnerkindl, line 164: That's what I suggested above, and it immediately follows from `NoFutureAdmission`.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+    \A n \in Nodes : processed[n] \subseteq {m.id : m \in network}
+
+\* I5: DECISION LOCALITY.
+\* If two nodes have both processed id, that id exists in network.
+\* Structural consequence: the gate depends only on (m.cslot, current_slot),
+\* not on n. Same (m.cslot, current_slot) => same decision at every node.
+DecisionLocalityOnly ==
```
muenchnerkindl, line 168: This one follows immediately from `NoPhantomProcess`.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+ASSUME NodesAssumption  == IsFiniteSet(Nodes)  /\ Nodes  # {}
+ASSUME MsgIDsAssumption == IsFiniteSet(MsgIDs)
+
+\* Node liveness states, named rather than written as bare strings.
+Up   == "up"
+Down == "down"
+
+VARIABLES
+    \* @type: Int;
+    current_slot,
```
muenchnerkindl, line 48: NB: This is a global variable, meaning that the spec assumes that all nodes have a perfect view of the current time (slot). But given that time is only used as an upper bound, the mechanism could be weakened so that `current_slot` models the ideal time but that each node has its own time reading `local_slot[n]` that may lag behind `current_slot` and is updated by an action that either copies `current_slot` into `local_slot[n]` or increases `local_slot[n]`, bounded by `current_slot`.

specifications/VortexDSE/Vortex_DSE_CSlot.tla
```diff
@@ -0,0 +1,242 @@
...
+    \* @type: Set({ id: Str, cslot: Int });
+    network,             \* in-flight messages (SET = no ordering)
```
muenchnerkindl: NB: "in-flight" in the sense of messages having been sent at any point: the spec never discards messages

specifications/VortexDSE/Vortex_DSE_CSlot_AE.tla
```diff
@@ -0,0 +1,269 @@
...
+\* Models assumption A3 (in-spec loss envelope).
+Reconcile ==
+    /\ \A n \in Nodes : phase[n] = Frozen
+    /\ LET union_view == UNION { processed[n] : n \in Nodes }
+       IN committed_set' = [n \in Nodes |-> union_view]
+    /\ phase' = [n \in Nodes |-> Committed]
+    /\ UNCHANGED <<current_slot, network, processed>>
+
+\* Adversarial duplicate / replay injection.
+DuplicateInject(id, fake_cslot) ==
```
muenchnerkindl: Again, this subsumes `Submit`.

specifications/VortexDSE/Vortex_DSE_CSlot_AE.tla
```diff
@@ -0,0 +1,269 @@
...
+\* m.cslot = current_slot at admission time.)
+NoReorderAcrossCslot ==
+    \A n \in Nodes : \A id \in processed[n] :
+        \E m \in network : m.id = id /\ m.cslot = current_slot
+
+\* AE-I5: PHASE PROGRESSION VALID.
+\* A node's phase is always one of the three legal states. Forward-only
+\* transitions are enforced structurally by the Freeze, Reconcile, NextCslot
+\* guards; declared here as an explicit type-level safety net.
+PhaseProgressionValid ==
```
muenchnerkindl, line 227: That's a direct consequence of the type invariant,

specifications/VortexDSE/Vortex_DSE_CSlot_AE.tla
```diff
@@ -0,0 +1,269 @@
...
+\* view. (This rules out the "frozen and then dropped" failure mode.)
+CommittedSupersetsProcessed ==
+    \A n \in Nodes :
+        phase[n] = Committed => processed[n] \subseteq committed_set[n]
+
+\* AE-I3: NO PHANTOM IN COMMITTED.
+\* Every id in any committed_set corresponds to a real network record with
+\* m.cslot = current_slot. AE cannot fabricate messages, only union real
+\* admissions.
+NoPhantomInCommitted ==
```
muenchnerkindl, line 214: The three invariants claimed here follow from
``\A n \in Nodes : processed[n] \subseteq {m.id : m \in {mm \in Network : mm.cslot = current_slot}}``
and
``\A n \in Nodes : phase[n] = Committed => committed_set[n] = UNION {processed[nn] : nn \in Nodes}``
(These two formulas can be made easier to read by introducing intermediate definitions.)

specifications/VortexDSE/Vortex_DSE_CSlot_AE.tla
```diff
@@ -0,0 +1,269 @@
...
+(*                                                                          *)
+(* Fairness assignment:                                                     *)
+(*  - SF(Reconcile): once all nodes are frozen, AE must run.               *)
+(*  - SF(NextCslot): once all nodes are committed, slot must advance.      *)
+(*  - WF(Freeze(n)) per node: each node eventually freezes.                *)
+(*  - NO fairness on Process / Submit / DuplicateInject (same rationale as *)
+(*    core module: late delivery is dropped by design; adversary unfair).  *)
+(***************************************************************************)
+
+Fairness ==
```
muenchnerkindl, line 242: I believe that weak fairness for `Reconcile` and `NextCslot` should be enough because once enabled, these actions only become disabled by taking them.

specifications/VortexDSE/Vortex_DSE_CSlot_AE.tla
```diff
@@ -0,0 +1,269 @@
...
+\* AE-L1: EVENTUAL COMMIT.
+\* Every node eventually commits for the cslot it participates in.
+EventualCommit ==
+    \A n \in Nodes :
+        (phase[n] = Open) ~> (phase[n] = Committed)
+
+\* AE-L2: EVENTUAL AGREEMENT.
+\* If two nodes both reach the committed phase, MerkleAgreement holds.
+\* (Safety + liveness composition.)
+EventualAgreement ==
```
muenchnerkindl, line 258: As you say, that is a direct consequence of `EventualCommit` and `MerkleAgreement`, but perhaps it is worthwhile stating it as an explicit property.

specifications/VortexDSE/Vortex_DSE_CSlot_AE_Proofs.tla
```diff
@@ -0,0 +1,58 @@
...
+EXTENDS Vortex_DSE_CSlot_AE, TLAPS
+
+
+-------------------------------------------------------------------------------
+(*                  PART A — TYPE INVARIANT                                 *)
+
+LEMMA InitType == Init => TypeInvariant
+  BY DEF Init, TypeInvariant, MsgRecord
+
+LEMMA NextType == TypeInvariant /\ [Next]_vars => TypeInvariant'
```
muenchnerkindl, line 17: Would a one-step proof work here, i.e.
``BY DEF TypeInvariant, MsgRecord, vars, Submit, ...``
?

specifications/VortexDSE/Vortex_DSE_CSlot_ExactlyOnce_Proof.tla
```diff
@@ -0,0 +1,295 @@
...
+(***************************************************************************)
+
+\* NOTE: TypeInvariant is REQUIRED as a hypothesis here. The Process(n,m) case
+\* must conclude mm.id \in MsgIDs from mm \in network, which holds only because
+\* network \subseteq MsgRecord — a TypeInvariant conjunct. Earlier this lemma
+\* unfolded TypeInvariant via USE DEF but never ASSUMED it, so that fact was
+\* not in scope and the mm.id \in MsgIDs obligation failed silently (tlapm does
+\* not return a non-zero exit code on unproved obligations). TypeInvariant is
+\* discharged in the theorem below via the machine-checked TypeCorrect.
+LEMMA NextStrictExactlyOnce ==
```
muenchnerkindl, line 104: There is nothing wrong about these proofs, but I strongly suspect that they could be collapsed to a few lines. We have often observed that Claude tends to produce verbose proofs.

specifications/VortexDSE/Vortex_DSE_CSlot_ExactlyOnce_Proof.tla
```diff
@@ -0,0 +1,295 @@
...
+        BY <1>6 DEF Tick
+  <1>7. CASE UNCHANGED vars
+        BY <1>7
+  <1>8. QED
+        BY <1>1, <1>2, <1>3, <1>4, <1>5, <1>6, <1>7 DEF Next
+
+\* We carry TypeInvariant /\ StrictExactlyOnceInv as ONE inductive invariant so
+\* the Process case always has network \subseteq MsgRecord in scope. Projecting
+\* onto the StrictExactlyOnce conjunct gives the exported safety theorem.
+THEOREM StrictExactlyOnceCorrect == Spec => []StrictExactlyOnce
```
muenchnerkindl, line 118: NB: A useful pattern when proving two invariants in succession, i.e.
``Spec => []A``
``Spec => []B``
where in the proof of the second invariant you want to use invariant `A` is to set up the second proof as follows:
```
THEOREM Spec => []B
<1>1. Init => B
<1>2. A /\ B /\ [Next]_vars => B'
<1>. QED  BY <1>1, <1>2, InvA, PTL DEF Spec
```
where `InvA` is the theorem asserting the first invariant. That is, you introduce the first invariant as an extra hypothesis in the preservation step of the second one (you may even introduce `A'` if useful), and this is justified by the theorem stating the first invariant. In particular, this avoids having to restate the invariant as the conjunction of `A` and `B`, to factor out the preservation step for invariant `A` as a separate lemma, and appealing to it in the proof of the second invariant.

specifications/VortexDSE/Vortex_DSE_CSlot_Proofs.tla
```diff
@@ -0,0 +1,263 @@
...
+
+
+-------------------------------------------------------------------------------
+(*                  PART A — TYPE INVARIANT (type-correctness)             *)
+
+\* (1) The initial state satisfies the type invariant.
+LEMMA InitType == Init => TypeInvariant
+  BY DEF Init, TypeInvariant, MsgRecord
+
+\* (2) Every step (or stutter) preserves the type invariant.
```
muenchnerkindl, line 44: Similar comments apply here, but as long as the proofs are checked, that's fine.

specifications/VortexDSE/Vortex_DSE_CSlot_Skew.tla
```diff
@@ -0,0 +1,152 @@
...
+VARIABLES
+    node_slot,    \* [Nodes -> Int]  per-node clock
+    network,      \* set of msg records
+    processed,
+    persisted,
+    node_state
+
+vars == <<node_slot, network, processed, persisted, node_state>>
+
+MsgRecord == [id: MsgIDs, cslot: Nat, origin: Nodes]
```
muenchnerkindl: Where is the additional `origin` field used?

specifications/VortexDSE/Vortex_DSE_CSlot_Skew.tla
```diff
@@ -0,0 +1,152 @@
...
+    \/ \E n \in Nodes : SkewedTick(n)
+
+Spec == Init /\ [][Next]_vars
+
+-------------------------------------------------------------------------------
+(*                          INVARIANTS                                       *)
+
+TypeInvariant ==
+    /\ node_slot   \in [Nodes -> Nat]
+    /\ \A m \in network : m.id \in MsgIDs /\ m.cslot \in Nat /\ m.origin \in Nodes
```
muenchnerkindl: This should rather be stated as `network \in SUBSET MsgRecord`. In fact, the semantics of TLA+ is such that this is not implied by the predicate asserted here.

specifications/VortexDSE/Vortex_DSE_CSlot_Skew.tla
```diff
@@ -0,0 +1,152 @@
...
+    /\ node_state  \in [Nodes -> {Up, Down}]
+
+\* The Tick guard guarantees this; it is asserted as invariant to make
+\* the skew bound an explicit, machine-checked property.
+BoundedSkew ==
+    \A n1, n2 \in Nodes :
+        /\ node_slot[n1] - node_slot[n2] <= MaxSkew
+        /\ node_slot[n2] - node_slot[n1] <= MaxSkew
+
+ExactlyOncePerNode ==
```
muenchnerkindl, line 132: Cf. my comment to `Vortex_DSE_CSlot.tla`: this one is a direct consequence of the type invariant.

specifications/VortexDSE/Vortex_DSE_CSlot_Skew.tla
```diff
@@ -0,0 +1,152 @@
...
+\* Local admission: a processed id has a network record whose cslot
+\* is at most the node's current local slot. (Strict equality holds
+\* at admission time; monotone clock means cslot <= node_slot[n] later.)
+CSlotLocalAdmission ==
+    \A n \in Nodes : \A id \in processed[n] :
+        \E m \in network : m.id = id /\ m.cslot <= node_slot[n]
+
+PersistedReflectsReality ==
+    \A n \in Nodes :
+        node_state[n] = Down =>
```
muenchnerkindl: Again, probably no need for this condition.

specifications/VortexDSE/Vortex_DSE_CSlot_Skew.tla
```diff
@@ -0,0 +1,152 @@
...
+            persisted[n] \subseteq {m.id : m \in network}
+
+NoPhantomProcess ==
```
muenchnerkindl, line 146: Immediate consequence of `CSlotLocalAdmission`.

specifications/VortexDSE/Vortex_DSE_CSlot_TTL.tla
```diff
@@ -0,0 +1,243 @@
...
+\* current slot AND it has not already been processed. No window, no TTL.
+\* Late delivery (m.cslot < current_slot) => permanent reject.
+\* Future-dated (m.cslot > current_slot) => reject now; would only be
+\* admitted if the message is delivered when the slot matches.
+Process(n, m) ==
+    /\ n \in Nodes
+    /\ m \in network
+    /\ node_state[n] = Up
+    /\ m.id \notin processed[n]              \* exactly-once guard (local)
+    /\ m.cslot = current_slot                \* STRICT C-slot equality
```
muenchnerkindl, line 96: If I'm not mistaken, the only difference to `Vortex_DSE_CSlot.tla` is that here equality is required whereas the other spec has `<=`. Therefore, all safety properties verified for the other spec also hold for this one. Moreover, the only liveness properties stated here are independent of the processing mechanism (and indeed, liveness for sent messages would not hold here because a message received late will not be accepted anymore). So, I wonder if this spec is of real interest?

muenchnerkindl (commented):
I'm a little late to the party, but perhaps you'll find my comments helpful nevertheless.

vasilisnasopoulos:
All of it applied — thank you, this was a much closer read than I had any right to expect.

**The proofs.** You were right that they collapse. `NextType`, `NextSafe` and `NextStrictExactlyOnce` were case splits where one `BY DEF` does the job, and `StrictExactlyOnceCorrect` was carrying the type invariant through as a conjunct instead of appealing to `TypeCorrect`, which is exactly the pattern you described:

| | lines | obligations |
|---|---|---|
| `Vortex_DSE_CSlot_Proofs` | 263 → 88 | 191 → 23 |
| `Vortex_DSE_CSlot_ExactlyOnce_Proof` | 295 → 131 | 128 → 19 |
| `Vortex_DSE_CSlot_AE_Proofs` | 60 → 36 | 32 → 10 |

Seven times more obligations than the argument needed.

**`Submit` and `DuplicateInject` are now one `Send`.** They differed only in preconditions nothing relied on, and with no fairness on either the split distinguished nothing — as you say.

**The invariant sections.** Each module now names the property of interest and marks the rest as following from it, rather than presenting a flat list that reads as more strength than is there. In the agreement layer I took the two facts you gave — `ProcessedAreCurrentSlot` and `CommittedIsUnion` — as the pair everything else derives from, with intermediate definitions as suggested. The corollaries stay defined and checked, but as regressions rather than claims.

**`PersistedReflectsReality`** — no reason at all; it holds unconditionally, verified, and the guard is gone in all three modules.

**Skew.** `network \in SUBSET MsgRecord` now, and thank you for being precise about why the quantified form does not imply it. The `origin` field was written and never read, so an adversary spoofing it changed nothing observable — it is removed rather than left looking like modelled adversarial power.

**AE fairness** — weak is enough for both, confirmed against the liveness model.

On the global `current_slot`: `Vortex_DSE_CSlot_Skew` is that weakening, with `node_slot[n]` per node and a bounded-drift tick. It is a sibling module rather than a refinement of the core, which is probably why it did not read as an answer to the point.

Seven TLC models and three proof modules, all green locally.

vasilisnasopoulos:
You are right on both counts, and as written the module did not earn its place. Rather than argue for it, I have made it say the two things you pointed at.

**The relationship is now proved, not asserted.** `Vortex_DSE_CSlot_TTL_Proofs` establishes `Spec => C!Spec`: the strict mode refines the default one under the identity mapping, so every safety property of the default is inherited and none of it is reproved. Your observation that equality is the only difference is exactly what the proof turns on — and the type invariant is needed in precisely one place, going from `m.cslot = current_slot` to `m.cslot <= current_slot`, since that step needs both to be naturals. 34 obligations.

**And the loss is now checked.** `MC_Vortex_DSE_CSlot_TTL_admission.cfg` states `EventualAdmission` and registers it as a liveness failure, which is the point of the module: a message whose slot has passed is refused for good. That property holds under the default rule and does not hold here, so the difference between the two modes is a machine-checked fact rather than a sentence in a comment.

That, I think, is the only reason for the module to exist: not that the strict rule is interesting on its own, but that the pair makes the trade-off explicit. The default admits late messages and keeps eventual admission; the strict rule bounds memory and gives it up. Neither dominates, and now neither claim rests on prose.

If you would still rather the directory carried one admission rule, I am happy to drop it — the refinement proof would then move into the core module's README as a remark.

muenchnerkindl:
Thank you, LGTM.

specifications/VortexDSE/README.md
```diff
@@ -0,0 +1,96 @@
+# Vortex DSE — C-slot admission and per-slot agreement
+
+Vortex DSE is a deterministic consensus protocol in which a message carries its
```
lemmy: "deterministic" doesn't seem to be needed here.

specifications/VortexDSE/README.md
```diff
@@ -0,0 +1,96 @@
...
+own slot stamp and each node decides admission locally, against its own clock.
+There is no leader, no quorum and no vote: admission is an O(1) local predicate,
+and cross-node agreement on the per-slot input set is established afterwards by
+a separate layer.
+
+These specifications model that structure. They are the formal counterpart of a
```
lemmy: What is “that structure”? Is it the “separate layer” or the part that has a very high-level characterization above?

In general, the README doesn’t help readers understand Vortex DSE’s conceptual functioning. Most of the other examples in this repository are accompanied by primarily academic publications. It would be beneficial for this example to also state the problem and how the solution works. The TLA+ specifications provide the formal proof that the solution solves the problem.

specifications/VortexDSE/README.md
```diff
@@ -0,0 +1,96 @@
...
+## Scope
+
+`Vortex_DSE_CSlot_Skew` bounds pairwise clock skew structurally, by forbidding
+any tick that would breach `MaxSkew`. It states the assumption; it does not
+model the mechanism that maintains it. Likewise `Reconcile` is a single atomic
+step at specification level — the multi-round protocol underneath is out of
+scope here.
+
+Source repositories, including the whitepaper and the model-checking logs:
+<https://github.com/vasilisnasopoulos>
```
lemmy: Why not deep links?

lemmy:
Since I lack the time to familiarize myself with Vortex DSE, I sought a review from a frontier model. Here’s the unfiltered response:

> **AI-based assessment:** The specs model a meaningful but narrow component: local time-slot admission, duplicate filtering, abstract crash/rejoin state, and the difference between late-tolerant and strict admission. They do **not** yet model a meaningful consensus system. The project’s own public documentation similarly describes C-slot as a local admission predicate rather than consensus and says the admission-to-agreement system is not formally composed. 
>
> The core uses one globally visible `network` set and records only processed message IDs. It therefore does not represent separate per-node delivery state, selective loss or equivocation, conflicting payloads, within-slot ordering, or replicated application state. The agreement module’s `Reconcile` action atomically assigns the union of every node’s processed set to all nodes, so agreement is assumed by the action rather than established by a distributed reconciliation protocol. 
>
> The TTL model also never deletes network, processed, or persisted state, so its bounded-memory motivation is not actually modeled. 
>
> **Recommendation:** include it only if it is framed as **slot admission plus idealized set reconciliation**. Keeping the **consensus protocol** framing should require a composed model with per-node channels, loss/equivocation and crashes during reconciliation, persistent ordered history or state-machine effects, garbage collection, and end-to-end agreement, validity, and termination properties.

vasilisnasopoulos:
You are right on every point, and the framing you quote is better than mine.

**Where the overstatement came from.** The specifications in this directory are one part of a much larger private set — a bit over a hundred modules and around a hundred and twenty model configurations, which between them cover the things you list as missing: per-node clocks and bounded skew, lossy delivery, equivocation and accountability, crash and rejoin composed with agreement, the timing layer that keeps the slot boundary. I chose to publish this part and keep the rest private, and then wrote the README as though it described the whole thing. That is how "consensus" ended up in the first sentence of a directory that does not model consensus.

I am not offering the private material as evidence — you cannot review what you cannot read, and a reviewer should not be asked to take a claim on trust. I mention it only because it explains the mistake, and because it is the honest answer to "what is that structure": it is the structure of a system most of which is not here.

**So, for what is here:** slot admission plus idealized set reconciliation. The README will say that and nothing larger.

The specific points all hold:

- One global `network` set. No per-node delivery state, no selective loss, no conflicting payloads for one id, no within-slot ordering, no replicated application state.
- `Reconcile` assigns the union in a single atomic step. Agreement is a property of that action, not something a reconciliation protocol establishes. I wrote as much in the scope note and then titled the module an agreement layer, which reads as more than it is.
- The TTL module never deletes anything, so bounded memory — the whole reason that mode exists — is not modelled. I added a liveness failure showing what the strict rule refuses, and left unmodelled the half that motivates it. That is the weakest thing I submitted.

On "that structure": the sentence pointed at nothing definite, and it is going.

On the README more generally — you are right that the other examples come with a paper and this one does not. I will rewrite it to state the problem, how the solution works conceptually, and explicitly what these modules do not cover, so a reader can tell what is proved and what is left open. I would rather that be plain than have someone infer a stronger result from one word.

vasilisnasopoulos:
The README is rewritten.

It opens with the problem — coordination costs rounds, and a round cannot outrun the signal between the parties — and then narrows to the question these modules actually answer: whether a node can decide admission without asking anyone. Agreement on the resulting sets is named as a separate question rather than folded into the first, which is where the earlier version went wrong.

There is a section for what is not covered, in your terms: not a consensus protocol; one global `network` set with no per-node delivery, selective loss, conflicting payloads or within-slot ordering; `Reconcile` as a single atomic step rather than a protocol; the TTL module never deleting, so the bounded memory that motivates it is unmodelled; `Skew` stating an assumption rather than a mechanism; admission and agreement never composed.

On the missing publication — the whitepaper is now linked. The specifications also belong to a larger body of work that is not public, and the README says so as context for why these pieces look narrow, explicitly not as evidence: nothing here depends on a module that is absent.

